Issue #44, opened 2018-02-14
Multiple imports of random in pma_scan module


shaddygarg:
The file engine.py in pma_scan module has multiple imports of random https://github.com/viraintel/OWASP-Nettacker/blob/master/lib/scan/pma/engine.py#L10 and https://github.com/viraintel/OWASP-Nettacker/blob/master/lib/scan/pma/engine.py#L12
_________________
**OS**: `Ubuntu`

**OS Version**: `17.10`

**Python Version**: `2.7`

Ali-Razmjoo:
Hello,

Thank you for sharing issue, I'll check it soon.

Regards.

Ali-Razmjoo:
Hey,

I've seen it's importing double, so let me close this issue and check your PR.

Best Regards.

PR #45, opened 2018-02-14
Fixes multiple import issue in pma_scan

shaddygarg:
This PR is in reference to the issue #44.
_________________
**OS**: `Ubuntu`

**OS Version**: `17.10`

**Python Version**: `2.7`

### DIFF
--- a/lib/scan/pma/engine.py
+++ b/lib/scan/pma/engine.py
@@ -7,7 +7,6 @@ import time
 import json
 import threading
 import string
-import random
 import requests
 import random
 import os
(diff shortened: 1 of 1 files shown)


Ali-Razmjoo:
Hello,


thank you for your contribution, I'll review and get back to you soon.

Best regards.

Ali-Razmjoo:
Hello,

I've checked your PR, the second `break` in the code was for the `for` loop but it seems I implemented it wrong! two `break`s are not working after each other, this PR also need to be fixed.

I've found a solution on [StackOverFlow](https://stackoverflow.com/questions/653509/breaking-out-of-nested-loops) and [here](https://www.python.org/dev/peps/pep-3136/). please implement a new solution to break the `for` loop in the `while` loop. let me know if you want any help.

Best Regards.

shaddygarg:
Hi, 
Thanks for the quick review. I saw the solution on StackOverFlow. It is fascinating to know but it doesn't consider the case when internal break is caused by the else block. Therefore, I implemented it using a variable keyboard_interrupt_flag and an if-else block. Let me know any further improvements!
Thanks and regards.

Ali-Razmjoo:
Hello @shaddygarg,

Sorry for the delay in responding and thank you for your contribution. I committed a simple[[1]](https://github.com/viraintel/OWASP-Nettacker/commit/578274699dc64b15c0997eab37c11d8e30b74f56) change.

I hope we see more contributions from you in future.

Best Regards.

PR #54, opened 2018-02-23
Admin scanner

Tikam02:
# admin scanner - /scan/admin/engine.py

It scans all admins pages including .html ,.asp , sysadmins and in future i'll more admin directories.
there is already pma for phpadmin sites but not for others indexes so i created  a separate module for admin sites.
screenshot from 2018-02-23 12-33-10


Ali-Razmjoo:
Hello,

Thank you for your contribution, we are looking forward to seeing more contribution from you.

Best regards.

PR #96, opened 2018-03-28
Password List Generator

pradeepjairamani:
https://github.com/pradeepjairamani/password_list_generator
Implemented password list generator in lib/payload/password_list_generator.py
Usage:

if you want to output the generated list in a file:

> from lib.payload.password_list_generator import generate
> password_list = generate("word_filename.txt")

if you don't want to output the generated list in a file:

> from lib.payload.password_list_generator import generate
> password_list = generate(None)

2. Implemented Joomla Version Scanner

The joomla version scanner scans for what joomla version the site is using using joomla.xml file 
_________________

**OS**: `Linux`

**OS Version**: `Ubuntu 16.04`

**Python Version**: `2.7.6, 3.4.3`


Ali-Razmjoo:
Hey,

Thanks for your contribution, a few changes need to be more handy with the framework.

* relocate the generator with `lib/payload/password_generator/engine.py` (use `engine.py` like other modules) and create its own `readme.md` file.
* do not use any `raw_input` or `input`, I should maintain the automating process in the framework for the reason.
* `generate("file.txt")` or `generate(None)` could be optimized. e.g. if user need to create a file it could be `generate(filename="file.txt")` otherwise `generate()` could return an array with passwords. in this case, you may modify the module like this. `def generate(filename=None):`. later we can add other methods to generate the password. e.g
```python
def generate(filename=None, common=None, suffix=False, prefix=False, username=None):
# generate common passwords
# generate with the suffix
# generate with the prefix
# generate depend on the username (e.g. username="Ali", output: ["Ali1", "Ali123", "Ali123!", "Ali!@#", "Ali2018"])
```
* if you need to `print` something, please use `core.alert` 
* use `core.log.__log_into_file(filename, mode, data, language, final=True)` for writing files (notice: use `final=True`)
* apply PEP8 fix the lib
* do not forget to add `__init__.py` to the lib folder.
* add code based docuemntation.
* add the document into your readme (like you did) I will add them to wiki

#1 I should probably write these roles in the documents too.

Best Regards.

pradeepjairamani:
Ok, I will make the changes and
Please give me a little tip for taking input without input or raw_input

Best Regards
Pradeep Jairamani

Ali-Razmjoo:
Hi,
* Please give me a little tip for taking input without input or raw_input
you may use default values and create an input method for changing the defaults.

* secondly there is no need for the username as in the data is already being generated from user specified details like name, last name, email, etc.

my bad, I think using the default (like I said) would solve this.

let me know if you have any more questions.

Regards.

pradeepjairamani:
My password list generator takes input from the user for generating password list like first name, last name, dob, email, phone number, vehicle number for generating the password list for the user.
There will be no use of creating a module for generating password list which would end up with default values always, we can get a file from the internet for that

Best Regards
Pradeep Jairamani

Ali-Razmjoo:
could the function work like `def generate(username="razmjoo", phone_number="", birth_year="", email="", name="ali", last_name="razmjoo", etc..)`  ?

Regards.

pradeepjairamani:
Sure, sorry for the confusion and I will make the changes

Best Regards

pradeepjairamani:
Hello,
I am getting this error during using __log_into_file

  File "lib/payload/password_generator/engine.py", line 156, in generate
    __log_into_file(filename, 'w', json.dumps(unique_list), language)
  File "core/log.py", line 191, in __log_into_file
    submit_logs_to_db(language, data)
  File "api/__database.py", line 182, in submit_logs_to_db
    """.format(log["HOST"], log["TIME"], log["PORT"], log["TYPE"], log["CATEGORY"],

Now I can add the data using 
        data = json.dumps({'HOST': target, 'USERNAME': '', 'PASSWORD': '', 'PORT': port, 'TYPE': '',
                           'DESCRIPTION': json.dumps(unique_list), 'TIME': now(),
                           'CATEGORY': "payload",
                           'SCAN_ID': scan_id, 'SCAN_CMD': scan_cmd})

but there is no use of sending this data to the database and secondly, the password generator can generate up to 1 lakh values ( Depending on the data given by the user ) so saving it in the database won't be a good idea.

Best Regards

Ali-Razmjoo:
Hey,

I mention you to use `final=True` before.

* use `core.log.__log_into_file(filename, mode, data, language, final=True)` for writing files (notice: use `final=True`)

what is `1 lakh`? I cannot understand.

let me know if it fixes your problem.

Regards.

Ali-Razmjoo:
Hey,

Thanks for the upgrades and contribution. PR Merged.

Best Regards.

PR #112, opened 2018-04-23
More accurate for FILTERED Scan

VictorSuraj:
Describe your changes or your new module and let us know which OS and Python version you tested on.
_________________
**OS**: `Any'

**OS Version**: `x`

**Python Version**: `2.7`

### DIFF
--- a/lib/scan/port/engine.py
+++ b/lib/scan/port/engine.py
@@ -119,9 +119,16 @@ if "--method-args" in sys.argv and "port_scan_stealth" in " ".join(sys.argv).low
     WINDOWS = True
     conf.verb = 0
     conf.nofilter = 1
+def check_closed(ip):
+    for i in range(1,10):
+        s = sr1(IP(dst=ip)/TCP(dport=i),timeout=2,verbose=0)
+        if s!='SA' and s!=None:
+            return i
 
+global closed_port
 
 def filter_port(ip, port):
+    closed_port=check_closed(ip)
     s = sr1(IP(dst=str(ip)) / TCP(dport=port, flags='S'), timeout=2, verbose=0)
     try:
         if s != 'SA':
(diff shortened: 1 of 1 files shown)


Ali-Razmjoo:
Hey,

Thanks for your contribution, is this necessary to use `global closed_port`? this function defined in the same level of it called.

Regards.

VictorSuraj:
OK I am removing this .   I wanted to use it for fast scanning purpose. But now its no need.
I am going to update engine.py to remove this.

pradeepjairamani:
Hello,

If it is important to use  closed_port globally, you can also use
`closed_port = int()`

Best Regards

Ali-Razmjoo:
@VictorSuraj, Great,

To be sure, please patch the possible bug.

```python
def check_closed(ip):
    for i in range(1, 10):
        s = sr1(IP(dst=ip) / TCP(dport=i), timeout=2, verbose=0)
        if s != 'SA' and s != None: #if this never happen, function will return None
            return i
```
if in any case, the conditions not pass the `if`, it will never `return i`. so it's best to add `return 0` in end of function to prevent possible errors.

Regards.

VictorSuraj:
I review it, "global closed_port" is used inside function filter_port(ip,port).
I think it well used here. If "global closed_port" remove from function "filter_port( ip,port)" then outputs remains same.
So I think we should remove "global closed_port".

Ali-Razmjoo:
@VictorSuraj, yes, it's not necessary so you should remove it.

shaddygarg:
Hey,
There are four empty lines between two functions : https://github.com/zdresearch/OWASP-Nettacker/pull/112/files#diff-ce3bf9d902408140b4d7678a6395911bR131. You should make them two instead of four.
Regards.

VictorSuraj:
@Ali-Razmjoo 
No error comes out also in that rare  case if any server's initial ten ports open.
because   here our code check any closed port in initial ten port if any port is closed then it will return 
the port_number.

VictorSuraj:
I think we should change as a return 0 in the end if a rare case caused as ten initial port opened in a server 
then it will return None or 0.

Ali-Razmjoo:
@VictorSuraj, so please add `return 0` for the rare cases, also as @shaddygarg mentioned, please use PEP8 style (I can modify your code but it's best to use pycharm or sublime),

SriHarshaGajavalli:
@VictorSuraj 
I hope with all the feedback you have received, you would be updating this PR. I would also like you to make small changes in your PR. The field `os` in PR Template refers to the operating system that you're using and the version of it in the field `OS Version`.

PR #249, opened 2020-04-24
Added Cookie headers related vulnerability module

### LINKED ISSUE #241: Set-Cookie header related Vulnerabilities
The set-cookie header has some flags like secure, httponly, samesite. If these flags are not enable then it can help attacker to exploit the site in various ways. But if samesite cookie is enabled then the website is safe from CSRF attacks. Ex In this blog: https://www.acunetix.com/blog/articles/chrome-tightens-csrf-protection/ after adding samesite cookie which will prevent the csrf attacks. I will add a module for searching for these flags in Set-Cookie header which will in return gives whether the application is vulnerable or not.
_________________
**OS**: `Linux`

**OS Version**: `Ubuntu`

**Python Version**: `3.6.9`

aman566:
fixed #241 
I have added a new module for finding set-cookie header-based vulnerabilities.

image

#### Checklist

- OS: `Linux`
- OS Version: `Ubuntu`
- Python Version: `3.x`

### DIFF
--- a/lib/transactions/maltego/nettacker_transforms/src/nettacker_transforms/transforms/http_cookie_vuln_scan.py
+++ b/lib/transactions/maltego/nettacker_transforms/src/nettacker_transforms/transforms/http_cookie_vuln_scan.py
@@ -1,0 +1,49 @@
+import random
+
+from canari.maltego.transform import Transform
+from canari.maltego.entities import URL
+from canari.framework import EnableDebugWindow
+from common.entities import NettackerScan
+
+from lib.vuln.cookie.engine import start
+
+from database.db import __logs_by_scan_id as find_log
+
+__author__ = 'Shaddy Garg'
+__copyright__ = 'Copyright 2018, nettacker_transforms Project'
+__credits__ = []
+
+__license__ = 'GPLv3'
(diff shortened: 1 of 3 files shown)


0xInfection:
This is an interesting addition actually. Good choice @aman566.  :D 

However the checks could be improved well enough by introducing checks like `samesite=strict|lax|None`. Since the introduction of [IETF's Incrementally Better Cookies](https://tools.ietf.org/html/draft-west-cookie-incrementalism-00), if a site doesn't set the `samesite` attribute to a cookie, the intention of the cookie would be by default set of `samesite=lax`. Also, cookies with `SameSite=None` must also specify `Secure`, meaning they require a secure context. So addition of these checks would enhance the operation of this framework if not causing too false positives. 

> NOTE: Only checking for `samesite` would cause false positives since cookies without a SameSite attribute will be treated as `SameSite=Lax`, and this is being considered as a standard since Chrome 80 implemented it and Firefox had it from version 69 itself.

aman566:
Yeah!!! Thank you @0xInfection I was trying to do that but I don't know the code is breaking. Well I will try once again. Thank you for the feedback :)

PR #251, opened 2020-04-27
Added CSP validator domains module

aman566:
I have added a CSP domains extractor module inspired from https://cspvalidator.org/. It will help pentesters to grab CSP domains fast and find XSS vulnerabilities exploited using wrong CSP implementation.

image

#### Checklist

- OS: `Linux`
- OS Version: `Ubuntu`
- Python Version: `3.x`


lib/vuln/csp_validator/engine.py
```diff
@@ -0,0 +1,189 @@
...
+                    break
+            except KeyboardInterrupt:
+                break
+        subs = __csp_validator(target, port, timeout_sec, log_in_file, language, time_sleep,
+              thread_tmp_filename, socks_proxy, scan_id, scan_cmd)
+        if len(subs) is not 0:
+            info(messages(language, "len_subdomain_found").format(len(subs)))
+            for sub in subs:
+               
+                data = json.dumps({'HOST': target, 'USERNAME': '', 'PASSWORD': '', 'PORT': '', 'TYPE': 'CSP domains found!!',
```
securestep9, line 180: this is incorrect: TYPE': 'CSP domains found!!', should be the module/finding identifier e.g. 'csp_validator_scan'  - there is inconsistent use of module id throughout this proposed module e.g. 'content_security_policy_vuln', 'csp_validator_vuln' ,'csp_validator_scan' etc. 

The idea of the module is good, but implementation needs some work. Also no actual CSP vulnerability/policy validation actually happens - only subdomain extraction, if that was the intention of this module then it needs to be renames to something like 'csp_domain_extraction_scan'

aman566: Thank you @securestep9 for the suggestions. As per suggested, then this module should fall in /lib/scan directory instead of /lib/vuln. I moved it to scan. Also, I changed the module name to csp_domain_extraction_scan. Please review it and suggest any changes if any

Ali-Razmjoo:
this can't be supported anymore. modules are yaml #440

PR #267, opened 2020-05-10
Changed XML-RPC brute to make tool fast

aman566:
The current wp_xmlrpc_brute module was doing the direct brute force on the web application with very weak wordlist files. So To be honest, without doing the brute force what we can do make tool faster is just check that xmlrpc is enable or not. This can be achieved using two techniques:
1. Seeing the response to the request  : http://domain/xmlrpc.php and the response should contain the string: "XML-RPC server accepts POST requests only."
2. But the second is more accurate: Making an actual post request and see the response which tells which requests we can make using xml.
So what I did is I just changed the code to make sure that it gives the faster response by just telling the web is vulnerable to xmlrpc or not. This will help pentester in many aspects:
1. He has to input this command to test for xmlrpc:
`python3 nettacker.py -m wp_xmlrpc_brute -i domain.com
`
2. After our tool confirms that the web is vulnerable, he can immideately go to burp suite and start intruder for brute forcing with **better username, password list.** This will save his time.
I just tried and it took much time so I decieded to do this.
The results are:
image

@Ali-Razmjoo please review it. I think it should help tool to get fast.
#### Checklist

- OS: `Linux`
- OS Version: `Ubuntu`
- Python Version: `3.6.9`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,7 +3,8 @@
 
 #ignore IDE settings
 *.idea*
-
+#ignore vscode files
+.vscode
 #setup
 build/*
 dist/*
--- a/lib/brute/wp_xmlrpc/engine.py
+++ b/lib/brute/wp_xmlrpc/engine.py
@@ -26,17 +26,11 @@ from lib.scan.wp_theme import small_themes
 
 def extra_requirements_dict():
     return {
-        "wp_users": ["admin", "root", "test", "anonymous", "user", "support"],
-        "wp_passwds": ["admin", "root", "test", "anonymous", "user", "1", "12345",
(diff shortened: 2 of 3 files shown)


securestep9:
@aman566   This PR is incorrect as it fundamentally changes the intended functionality of the module. The idea of checking for XMLRPC being enable is good ,but it should be implemented as a new separate scan module (e.g. wp_xmlrpc_scan) . This module is a for a specific brute forcing using wp.getUsersBlogs which bypasses login restrictions

aman566:
Closing this PR as suggested.

PR #283, opened 2020-06-04
Added shodan scan module

### LINKED ISSUE #204: Implementing Shodan reconnaissance in OWASP-Nettacker
There is no way that current version of nettacker can search on shodan and do information gathering. I will add this feature such that shodan search will also get implemented in nettacker.

aman566:
Fixed #204 
The api key switch for shodan is added. the folllowing command will search in the shodan database for results. The current versioni of this scan module searches for domain Name, IP, CIDR.

`python nettacker.py -i <Domain name> -m shodan_scan --method-args shodan_api_key=XXXXXX`
The shodan scan modules includes the following extraction:
1. ip+port.
2. Data upto 200 bytes.
3. Country name.
4. Organisation name.
5. cpe values.
6. hostname, if any.
7. CVEs and their cvss score.

Explanation of the working of **shodan_scan** module:
This module will run by adding extra switch **--method-args**.
For ex:
1. **python nettacker.py -i pintasuper.com -m shodan_scan --method-args "shodan_api_key=XXX&shodan_query_override=country:in port:6443"**
In this example the nettacker will find all the open kubernetes services in india.
2. If no shodan_query_override is given in the --method-args option then shodan will run on **-i** switch's target. For ex:
**python nettacker.py -i pintasuper.com -m shodan_scan --method-args "shodan_api_key=XXX"**
In this case shodan will run on pintasuper.com
3. It will work in case of CIDR too. as the internal process of nettacker divides the CIDR to ip range and iterate on it. So shodan will give it's results according to the CIDR entered. Ex:
**python nettacker.py -i 10.0.0.0/8 -m shodan_scan --method-args "shodan_api_key=XXX&shodan_query_override=country:in port:8080**" will search for all tomacat servers running in india in the CIDR range of 10.0.0.0/8.

The results in nettacker will vary according to the given API(Different shodan API plans will differ the results in nettacker).

Note: **shodan_api_key** and **shodan_query_override** are to be given to the tool for proper shodan working.

#### Checklist

- [X] Added the shodan module.
- OS: `Linux`
- OS Version: `Ubuntu 18.04`
- Python Version: `3.6.9`


core/targets.py
```diff
@@ -61,7 +61,7 @@ def target_type(target):
 
 
 def analysis(targets, check_ranges, check_subdomains, subs_temp, range_temp, log_in_file, time_sleep,
-             language, verbose_level, retries, socks_proxy, enumerate_flag):
+             language, verbose_level, retries, socks_proxy, enumerate_flag, scan_method):
```
securestep9, line 64: do not modify core for one module

core/targets.py
```diff
@@ -105,17 +105,21 @@ def analysis(targets, check_ranges, check_subdomains, subs_temp, range_temp, log
...
-            if not enumerate_flag:
-                info(messages(language, "checking").format(target))
-            if type(IPs) == netaddr.ip.IPNetwork:
-                for IPm in IPs:
-                    yield IPm
-            elif type(IPs) == list:
-                for IPm in IPs:
-                    for IP in IPm:
-                        yield IP
+            if("shodan_scan" in scan_method and "/" in target):
```
securestep9, line 110: do not modify the core for one module - this creates a hard link between the core and shodan module - the modules should be independent

lib/scan/shodan/engine.py
```diff
@@ -0,0 +1,364 @@
...
+            extra_requirements["shodan_api_key"][0],
+            extra_requirements=extra_requirements,
+        )
+        count = 0
+        if len(result) == 0:
+            info(messages(language, "shodan_false"))
+        if len(result) != 0:
+            for parts in result:
+                if verbose_level > 2:
+                    info(messages(language, "shodan_true").format(parts))
```
securestep9, line 310: rename the message "shodan_results_found"

lib/scan/shodan/engine.py
```diff
@@ -0,0 +1,364 @@
...
+            data = (
+                json.dumps(
+                    {
+                        "HOST": target,
+                        "USERNAME": "",
+                        "PASSWORD": "",
+                        "PORT": "",
+                        "TYPE": "shodan_scan",
+                        "DESCRIPTION": messages(
+                            language, "shodan_true"
```
securestep9, line 343: "shodan_results_found"

lib/scan/shodan/engine.py
```diff
@@ -0,0 +1,364 @@
...
+            socks_proxy,
+            retries,
+            num,
+            total,
+            extra_requirements["shodan_api_key"][0],
+            extra_requirements=extra_requirements,
+        )
+        count = 0
+        if len(result) == 0:
+            info(messages(language, "shodan_false"))
```
securestep9, line 306: change to "shodan_results_not_found"

core/targets.py
aman566: Sir here's the problem:
The current tool working need -i for the target and in shodan scan case (CIDR) the target will be : ex: 10.0.0.0/8.
but the internal working of tool converts this CIDR to IP. But thanks to shodan that it supports CIDR cases too with the following api call:
**https://api.shodan.io/shodan/host/search?key=shodan_api_key&query="net: 10.0.0.0/8"**
So if we use the --method-args for CIDR like:
**python nettacker.py -i pintasuper.com -m shodan_scan --method-args "shodan_api_key=<apikey>&shodan_cidr=10.0.0.0/8"**
Then it makes no sense to use -i and giving domain name as we are not using that. But without -i the tool won't work. Well I will fix it. Just need some time. will try to resolve it ASAP.

lib/scan/shodan/engine.py
```diff
@@ -0,0 +1,364 @@
...
+    timeout_sec,
+    log_in_file,
+    time_sleep,
+    language,
+    verbose_level,
+    socks_proxy,
+    retries,
+    num,
+    total,
+    shodan_api_key,
```
securestep9, line 179: shodan_api_key is already in extra_requirements

lib/scan/shodan/engine.py
```diff
@@ -0,0 +1,364 @@
...
+from core.targets import target_type
+from core.targets import target_to_host
+from core._time import now
+from core.log import __log_into_file
+import shodan
+
+
+def extra_requirements_dict():
+    return {
+        "shodan_scan": ["True"],
```
securestep9, line 22: why shodan_scan=True is needed here? seems unnecessary . Should be something  like "`shodan_query_override`" - if defined it will be used instead of target

lib/scan/shodan/engine.py
```diff
@@ -0,0 +1,364 @@
...
+        "status_code": [200, 401, 403],
```
securestep9, line 23: not sure about these - please explain why these status codes are needed and a scenario for a user to use them?

lib/scan/shodan/engine.py
```diff
@@ -0,0 +1,364 @@
...
+        "shodan_api_key": ["testkey"],
```
securestep9, line 24: better dummy value instead of "`testkey`" would be something like "`your_shodan_api_key_here`"

core/targets.py
securestep9: I think it should be like this:
**python nettacker.py -i pintasuper.com -m shodan_scan --method-args "shodan_api_key=XXXYYYYY&shodan_query_override=net:10.0.0.0/8"**

in this case `shodan_query_override` would be used and the target would be ignored.

if `shodan_query_override` method arg is defined it will allow the module to run some cool Shodan queries like this:

**python nettacker.py -i dummy.com -m  shodan_scan --method-args "shodan_api_key=XXXYYYY&shodan_query_override=country:in+port:6443"**  - this will return all instances of Kubernetes running in India (no way to define this using normal nettacker syntax"


Don't forget that Nettacker allows users to chain modules, so for example a user might want to do a subdomain scan and a shodan scan at the same time  on a target:

**python nettacker.py -i pintasuper.com -m subdomain_scan,shodan_scan --method-args "shodan_api_key=XXXYYYYY"** - in this case both scan methods would run

aman566: This is great idea. Actually it also give user simple user interface and he can type his own query. We need not to mention for CIDR now. The query which he types will automatically run and nettacker will give results according to it :)

aman566: One doubt. If user not gives that shodan_query_override then what should be the default query that should be run?
It is normal hostname scan?

aman566:
Sir I am not getting why flake8 gives error in requirements.txt. Invalid syntax???

lib/language/messages_en.py
```diff
@@ -149,6 +149,12 @@ def all_messages():
             "all_usernames": "the usernames",
             "all_passwords": "the passwords",
             "timeout_seconds": "the timeout seconds",
+            "Invalid_shodan_api_key": "{0}",
+            "shodan_api_key": "Shodan api key for shodan recon",
+            "shodan_results_found": "Found results: {0}",
+            "shodan_results_not_found": "Didn't found anything in the shodan database",
```
securestep9, line 155: typo/grammar fix, please change to : "Didn't find anything in the Shodan database"

lib/language/messages_en.py
```diff
@@ -211,6 +217,7 @@ def all_messages():
             "http_auth_failed": "http basic authentication failed to {0}:{3} using {1}:{2}",
             "http_form_auth_success": "http form authentication success - host:{2}:{3}, user: {0}, pass:{1} found!",
             "http_form_auth_failed": "http form authentication failed to {0}:{3} using {1}:{2}",
+            "using_shodan_query_override": "Using {0} as a target",
```
securestep9, line 220: Please change to: "Using {0} as a target for Shodan"

PR #369, opened 2021-01-17
Fixed some false positives in dir_scan module

### LINKED ISSUE #280: False Positives while running dir_scan
The current version of Nettacker scans for directory listing. But doing directory listing it searches for 401, 403, 200 status codes which is really a good approach. But in this manner what happens is many websites(90% where I used the tool) smartly creates some 404 pages where the status codes are 200 which is really bad for the tool giving false positives.
image

_________________
**OS**: `Linux`

**OS Version**: `Ubuntu 18.04`

**Python Version**: `3.6.9`

aman566:
Fixed #280 
Tried fixing dir_scan false positives by making a request to arbitary path which for sure gives 404 and storing the Content-Length of it. Using the difflib module, finding the similarity between the original request(From bruteforcing the path) and the request made for 404. If the ratio is greater than 0.95, then consider it as similar and skip it, else consider it

#### Checklist
- [x] I have followed the [Contributor Guidelines](https://github.com/zdresearch/OWASP-Nettacker/wiki/Developers#contribution-guidelines).

- OS: `Linux`
- OS Version: `Ubuntu 20.04`
- Python Version: `3.8`

### DIFF
--- a/lib/scan/dir/engine.py
+++ b/lib/scan/dir/engine.py
@@ -21,6 +21,7 @@ from core.compatible import version
 from lib.scan.dir import wordlist
 from lib.payload.wordlists import useragents
 import six
+from difflib import SequenceMatcher
 
 
 SESSION = requests.Session()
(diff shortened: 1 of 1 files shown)


lib/scan/dir/engine.py
```diff
@@ -179,6 +184,12 @@ def start(target, users, passwds, ports, timeout_sec, thread_number, num, total,
         trying = 0
         if target_type(target) != "HTTP":
             target = 'http://' + target
+        nowh3r3 = target + "/ThisFileIs404"
+        try:
+            r = SESSION.get(nowh3r3, verify=False, headers=user_agent, timeout=10)
```
securestep9, line 189: hardcoding 10 second timeout not right and it is too long - please use **timeout_sec** 
if user needs to increase the timeout they can always do that using command line options

securestep9 (approved):

Issue #371, opened 2021-01-20
False positives while running admin scan module


itsdivyanshjain:
This issue is similar to #280 but now we get false positive only while running admin_scan module
![admin-issues](https://user-images.githubusercontent.com/45870070/105126513-ed0d4a00-5b04-11eb-9cd0-03594dd84f75.jpg) 
_________________
**OS**: `Backbox`

**OS Version**: `7.1`

**Python Version**: `3.8`

itsdivyanshjain:
I like how previous false positive is tackled in dir_scan. So, i would like to take up this issue.

PR #372, opened 2021-01-20
Fixed Admin scan false positives

itsdivyanshjain:
#### Checklist

Bug #371 solved. The approach is same as PR #369 and this helps also to tackle admin_scan modules false positives.

- OS: `Backbox`
- OS Version: `7.1`
- Python Version: `3.8`


itsdivyanshjain:
@securestep9 should I just add `allow_redirects = False` in old code. 

But above solution will not help to tackle issues mentioned below:

- I have noticed that some site use encoding for url, means when we search for `example.com/path` it always redirects to some `example.com\somejiberrish`. 

- Another issue is that scan modules like `admin`, `cms`, `dir`, `drupal` (most scan module which search for specific path in the target) are also affected by such false positives.

Ali-Razmjoo:
this can't be supported anymore. modules are yaml #440

PR #391, opened 2021-02-13
Fix #390

### LINKED ISSUE #390: Improvement in csp_vuln mdule
Currently, the content_security_policy module only checks for presence for the `Content-Security-Policy` header. However, we should also check for the `<meta http-equiv="Content-Security-Policy"`  tag in the absence of the CSP header.
Moreover, we could also check for the `unsafe-inline` and `unsafe-eval` directives as to the weak implementation of CSP

ktg-123:
#### Checklist

I have made the improvement and done some code reformat

- OS: `Ubuntu`
- OS Version: `20.04`
- Python Version: `3.8.5`

### DIFF
--- a/lib/vuln/content_security_policy/engine.py
+++ b/lib/vuln/content_security_policy/engine.py
@@ -10,11 +10,7 @@ import threading
 import string
 import random
 import sys
-import struct
-import re
 import os
-from OpenSSL import crypto
-import ssl
 from core.alert import *
 from core.targets import target_type
 from core.targets import target_to_host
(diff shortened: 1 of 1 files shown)


securestep9:
@ktg-123 while a valid enhancement downloading and parsing the source of the page can add a significant time penalty. i propose that you add a method attribute for enabling this check and keep it on False by default, only enable if it is True. Check out how it is implemented in wp_scan_plugins module for example: https://github.com/OWASP/Nettacker/blob/9d41fb8a01c2e3903cc02e753bdf77b9332ac849/lib/scan/wp_plugin/engine.py#L32

ktg-123:
Thanks! I will check it out and make the changes

securestep9 (approved):

Issue #49, opened 2018-02-17
A better ICMP library


Ali-Razmjoo:
Hello,

I just notice that in #47 the `lib/icmp` and `icmp_scan` need to run as root to be working, I glad if someone can give us a better solution and contribute a new lib and replace the libraries.

let me know if anyone has any idea.

Best Regards.

shaddygarg:
Hey @Ali-Razmjoo, 
I read about the issue of running icmp scan as root but I can't seem to find a possible solution. This is because we need to create a raw socket while sending a icmp packet and linux doesn't allow creation of raw sockets without root privileges. Alternatively we can use the ping command to run in the background using os.system('ping host'). But it won't serve the same purpose as that of a full fledged icmp library. So, I think we should implement it in the following way.
1. Check if the user logged in has root privileges
2. If it has root permissions, use the existing library to perform the task
3. Otherwise we can resort to os.system('ping host')

Regards.

shaddygarg:
I would like to know your views on the above implementation.

Ali-Razmjoo:
Hello,

Please let me research on this topic, I will get back to you soon, thank you for sharing your solution.

Regards.

pallav12:
Hello, 
icmp_scannning is just a way to know if target is active, we can do this with simple ping like this
``
import os

hostname = "google.com" #example

response = os.system("ping -c 1 " + hostname)
if response == 0:

  print hostname, 'is up!'

else:

  print hostname, 'is down!'

``
am I missing something?

Ali-Razmjoo:
it gets extra time to do that, I think we are going to focus on docker image instead to avoid such issues.

securestep9:
@pallav12  the whole idea of Nettacker is to be 100% in python and not dependent on other tools installed in OS like nmap/ping etc-  calling an OS command is not scaleable in terms of pythons multi-threading. Most people run nettacker in Kali Linux as root anyway and can also use it in docker as @Ali-Razmjoo has already mentioned.

pallav12:
Ohh okay, makes sense

abinpaul1:
[icmplib](https://github.com/ValentinBELYN/icmplib) is a new library which provides a pure Python implementation of the ICMP protocol along with the option of running as non-root.

Ali-Razmjoo:
@abinpaul1 👍 

yeah we will add this as dependecy https://pypi.org/project/icmplib/

abinpaul1:
[...]

PR #397, opened 2021-03-05
Changed icmp_scan implementation to use icmplib library

abinpaul1:
#### Checklist
- [ ] The code is both Python 2 and Python 3 compatible.
- [x] I have added the relevant documentation.

This pull request contains the implementation of a new icmp library which provides an option to run ping as non-root user as specified in #49 . As I mentioned in the issue, icmplib is compatible only with python3.6+, if necessary I could modify the pull request to check for python version and decide on which code to use.

- OS: `Ubuntu`
- OS Version: `20.04`
- Python Version: `3.8.5`

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -22,3 +22,4 @@ six==1.15.0
 shodan==1.25.0
 pycurl==7.43.0.5
 xmltodict==0.12.0
+icmplib
--- a/core/attack.py
+++ b/core/attack.py
@@ -10,6 +10,7 @@ import sys
 import socks
 import socket
 import urllib3
+from icmplib import ping
 from core._die import __die_failure
 from core.alert import info
 from core.targets import target_type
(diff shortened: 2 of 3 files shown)


requirements.txt
```diff
@@ -22,3 +22,4 @@ six==1.15.0
 shodan==1.25.0
 pycurl==7.43.0.5
 xmltodict==0.12.0
+icmplib
```
itsdivyanshjain, line 25: Do mention the current version(or tested working version) of this package. So, dependbot push update of that package.

itsdivyanshjain (changes requested):
python2.7 docker builds is failing because icmplib support only python3

requirements.txt
itsdivyanshjain, line 25: mention condition to install icmplib if it using pip3 .

itsdivyanshjain: for `attack.py` and `engine.py` you should add condition that if user use python3 then this code run otherwise default old code should run(with root privileges). This will not break the program if user run this using python2.

abinpaul1:
> python2.7 docker builds is failing because icmplib support only python3

True. Do let me know how I should structure the modification. Should I do code modification like 
`
if sys.version_info > (3, 6):
    import icmplib
` 
to support two different codes based on python version.

itsdivyanshjain:
I think this tools is now going to support python3 only. So, wait @securestep9 response on this.

securestep9:
@abinpaul1 thanks for the PR , we will look at merging it a bit later - after we drop support for Python2. Currently there are a few bugs to fix before we can have a stable version which we can release for Python2 and then close it and move to Python3-only support

Ali-Razmjoo:
fixed #440

Issue #405, opened 2021-03-15
We need a module for MS Exchange CVE-2021-26855


securestep9:
With MSExchange hacking still making headlines every day since 4th March 2021 the attacks are going on all over the world with tens of thousands servers hacked he need a CVE-2021-26855 vuln detection module to help organisations using OWASP Nettacker to detect if they have any vulnerable Exchange servers in their networks.

Nmap script can be found here:

https://github.com/microsoft/CSS-Exchange/blob/main/Security/src/http-vuln-cve2021-26855.nse

we need to create a nettacker module for this task: **msexchange_cve_2021_26855_vuln**

ktg-123:
I would like to take up the issue

ktg-123:
The implementation of the module is the same as the map script. Please review

PR #406, opened 2021-03-15
Add CVE-2021-26855 vuln module

ktg-123:
#### Checklist

I have added the module as per #405 
- OS: `Ubuntu`
- OS Version: `20.04`
- Python Version: `3.8.5`

### DIFF
--- a/lib/vuln/msexchange_cve_2021_26855/engine.py
+++ b/lib/vuln/msexchange_cve_2021_26855/engine.py
@@ -1,0 +1,182 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
+import socket
+import socks
+import time
+import json
+import threading
+import string
+import random
+import sys
+import os
+from core.alert import *
+from core.targets import target_type
+from core.targets import target_to_host
+from core.load_modules import load_file_path
+from lib.socks_resolver.engine import getaddrinfo
+from core._time import now
+from core.log import __log_into_file
+import requests
+
+
+def extra_requirements_dict():
+    return {
+        "vuln_ports": [80, 443]
+    }
+
+
+def conn(targ, port, timeout_sec, socks_proxy):
+    try:
+        if socks_proxy is not None:
+            socks_version = socks.SOCKS5 if socks_proxy.startswith(
+                'socks5://') else socks.SOCKS4
+            socks_proxy = socks_proxy.rsplit('://')[1]
+            if '@' in socks_proxy:
+                socks_username = socks_proxy.rsplit(':')[0]
+                socks_password = socks_proxy.rsplit(':')[1].rsplit('@')[0]
+                socks.set_default_proxy(socks_version, str(socks_proxy.rsplit('@')[1].rsplit(':')[0]),
+                                        int(socks_proxy.rsplit(':')[-1]), username=socks_username,
+                                        password=socks_password)
+                socket.socket = socks.socksocket
+                socket.getaddrinfo = getaddrinfo
+            else:
+                socks.set_default_proxy(socks_version, str(socks_proxy.rsplit(':')[0]),
+                                        int(socks_proxy.rsplit(':')[1]))
+                socket.socket = socks.socksocket
+                socket.getaddrinfo = getaddrinfo()
+        s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+        sys.stdout.flush()
+        s.settimeout(timeout_sec)
+        s.connect((targ, port))
+        return s
+    except Exception as e:
+        return None
+
+
+def msexchange_vuln(target, port, timeout_sec, log_in_file, language, time_sleep,
+                   thread_tmp_filename, socks_proxy, scan_id, scan_cmd):
+    try:
+        s = conn(target, port, timeout_sec, socks_proxy)
+        if not s:
+            return False
+        else:
+            if target_type(target) != "HTTP" and port == 443:
+                target = 'https://' + target
+            if target_type(target) != "HTTP" and port == 80:
+                target = 'http://' + target
+            cookies = {
+                "X-AnonResource": "true",
+                "X-AnonResource-Backend": "localhost/ecp/default.flt?~3",
+                "X-BEResource": "localhost/owa/auth/logon.aspx?~3",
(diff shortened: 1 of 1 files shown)


securestep9:
hey @ktg-123 I tried it on many vulnerable servers searchable on Shodan, but the module failed to detect the vuln - please do some through testing and make sure that the module actually works. For references please do check out this Python scanner (it uses a slightly different detection method): https://github.com/KotSec/CVE-2021-26855-Scanner/blob/main/scan.py

and check the nuclei template here: 

https://github.com/projectdiscovery/nuclei-templates/blob/master/cves/2021/CVE-2021-26855.yaml

Important thing from nuclei is that we only need to check for responses with status_code 500 & 503 (this was not in MS nse script, but this speeds up the checking process)

lib/vuln/msexchange_cve_2021_26855/engine.py
```diff
@@ -0,0 +1,181 @@
...
+                target = 'http://' + target
+            cookies = {
+                "X-AnonResource": "true",
+                "X-AnonResource-Backend": "localhost/ecp/default.flt?~3",
+                "X-BEResource": "localhost/owa/auth/logon.aspx?~3",
+            }
+            if target.endswith("/"):
+                target = target[:-1]
+            path = '/owa/auth/x.js'
+            req = requests.get(target + path, cookies=cookies)
```
securestep9, line 76: @ktg-123  you must add verify=False here and timeout=timeout_sec
as we are connecting on https://ip/owa instead of https://fqdn verification will fail and the request fails silently as you simply kill the exception
Also you should add headers=useragent and get a random useragent from the useragent libraries (check @aman566 's merged PRs) otherwise the request will go out with python/requests as useragent and can be blocked

lib/vuln/msexchange_cve_2021_26855/engine.py
```diff
@@ -0,0 +1,181 @@
...
+            try:
+                global header_server
```
securestep9, line 78: no need for a global variable here

lib/vuln/msexchange_cve_2021_26855/engine.py
```diff
@@ -0,0 +1,181 @@
...
+                header_server = req.headers['x-calculatedbetarget']
+                if header_server == 'localhost':
```
securestep9, line 80: this is actually should be changed to `'localhost' in req.headers['x-calculatedbetarget']`

lib/vuln/msexchange_cve_2021_26855/engine.py
```diff
@@ -0,0 +1,181 @@
...
+                "X-AnonResource-Backend": "localhost/ecp/default.flt?~3",
+                "X-BEResource": "localhost/owa/auth/logon.aspx?~3",
+            }
+            if target.endswith("/"):
+                target = target[:-1]
+            path = '/owa/auth/x.js'
+            req = requests.get(target + path, cookies=cookies)
+            try:
+                global header_server
+                header_server = req.headers['x-calculatedbetarget']
```
securestep9, line 79: remove this line please - extra global var not necessary

lib/vuln/msexchange_cve_2021_26855/engine.py
```diff
@@ -0,0 +1,181 @@
...
+        # some error warning
+        return False
+
+
+def __msexchange_vuln(target, port, timeout_sec, log_in_file, language, time_sleep,
+                     thread_tmp_filename, socks_proxy, scan_id, scan_cmd):
+    if msexchange_vuln(target, port, timeout_sec, log_in_file, language, time_sleep,
+                      thread_tmp_filename, socks_proxy, scan_id, scan_cmd):
+        info(messages(language, "target_vulnerable").format(target, port,
+                                                            'Exchange Server SSRF Vulnerability'))
```
securestep9, line 96: @ktg-123  please add the CVE number to the message (there are many SSRf vulns in MS Exchange) so it looks like this:
`'Exchange Server SSRF Vulnerability CVE-2021-26855'`

lib/vuln/msexchange_cve_2021_26855/engine.py
```diff
@@ -0,0 +1,181 @@
...
+        __log_into_file(thread_tmp_filename, 'w', '0', language)
+        data = json.dumps({'HOST': target, 'USERNAME': '', 'PASSWORD': '', 'PORT': port,
+                           'TYPE': 'msexchange_cve_2021_26855_vuln',
+                           'DESCRIPTION': messages(language, "vulnerable").format(
+                               'Exchange Server SSRF Vulnerability'), 'TIME': now(),
```
securestep9, line 101: please add CVE number to the message as per comment above

lib/vuln/msexchange_cve_2021_26855/engine.py
```diff
@@ -0,0 +1,181 @@
...
+            try:
+                global header_server
+                header_server = req.headers['x-calculatedbetarget']
+                if header_server == 'localhost':
+                    return True
+                else:
+                    return False
+            except:
+                return False
+    except Exception as e:
```
securestep9, line 86: need to log the exception here using warn

lib/vuln/msexchange_cve_2021_26855/engine.py
```diff
@@ -0,0 +1,181 @@
...
+            path = '/owa/auth/x.js'
+            req = requests.get(target + path, cookies=cookies)
+            try:
+                global header_server
+                header_server = req.headers['x-calculatedbetarget']
+                if header_server == 'localhost':
+                    return True
+                else:
+                    return False
+            except:
```
securestep9, line 84: please add `Exception e` to enable debug

securestep9 (changes requested):

lib/vuln/msexchange_cve_2021_26855/engine.py
securestep9: please add a check for status_code in [500,503] to speed up the process - we are only interested in these status codes

ktg-123:
@securestep9 I have made the changes as per the review.

securestep9 (approved):

PR #429, opened 2021-06-12
Changed files in lib/ to python3 and removed py2 compatibility

aman566:
The main aim of project is now to make it completely python3 compatible. I have changed lib/ folder and made it fully python3 compatible. Now, python2 compatibility is removed.
#### Checklist
- [ ] The code is both Python 2 and Python 3 compatible.

- OS: `Linux`
- OS Version: `Ubuntu 20.04`
- Python Version: `3.9.3`


lib/argparse/v2/__init__.py
```diff
@@ -1,3 +0,0 @@
-#!/usr/bin/env python
```
dhirensr, line 1: not sure if you want to delete the files that are currently not being used, because in future it can be used. what do you think @Ali-Razmjoo ?

lib/scan/admin/engine.py
```diff
@@ -88,8 +87,7 @@ def check(target, user_agent, timeout_sec, log_in_file, language, time_sleep, th
                         )
                     )
                     return 1
-        if version() == 3:
-            content = content.decode("utf8")
+        content = content.decode("utf8")
```
dhirensr, line 90: does this work? i thought you need to pass "utf-8"

lib/argparse/v2/__init__.py
aman566: I thought, we are removing v2 completely from the project, so removed everything related to python2.x. Also this file was used only on one location and so @itsdivyanshjain has also removed the use of this file from that location.
@Ali-Razmjoo please give your thoughts on it

lib/scan/admin/engine.py
aman566: That was a typo. Then too it was working
image

lib/scan/wp_plugin/engine.py
```diff
@@ -74,8 +73,7 @@ def check(target, user_agent, timeout_sec, log_in_file, language, time_sleep, th
                 if n == retries:
                     warn(messages(language, "http_connection_timeout").format(target))
                     return 1
-        if version() == 3:
-            content = content.decode('utf8')
+        content = content.decode('utf8')
```
dhirensr, line 76: please update the typo everywhere

Ali-Razmjoo (approved):
since we are going to yaml base module before release python3; let's merge the libs!

PR #593, opened 2022-09-24
replace requests with iohttp

Ali-Razmjoo:
Requests lib use a lot of memory and cause a crash. Performance in iohttp is x40 higher.


Ali-Razmjoo:
there is no http method "trace" same as requests lib, we need to implement that separately and that's why `modules/vuln/http_trace_enabled.yaml` is removed.

PR #608, opened 2022-10-02
Core ssl lib and ssl scan module

Ali-Razmjoo:
- [x] core_exec module
- [x] proxychain
- [x] weak cipher suits
- [x] insecure cipher suits
- [x] expired certificates
- [x] soon to expire certificates
- [x] wrong host
- [x] self signed
- [x] untrusted root
- [x] revoked
- [x] pinning test
- [x] no-common-name
- [x] no subject
- [x] incomplete chain
- [x] cert info gathering

untrusted cert error codes all of them are supported: https://www.gnutls.org/manual/html_node/Error-codes.html

### DIFF
--- a/config.py
+++ b/config.py
@@ -118,7 +118,7 @@ def nettacker_user_application_config():
         "scan_ip_range": False,
         "scan_subdomains": False,
         "skip_service_discovery": False,
-        "thread_per_host": 100,
+        "thread_per_host": 1024,
         "parallel_module_scan": 1,
         "socks_proxy": None,
         "retries": 1,
--- a/core/module_protocols/core_exec.py
+++ b/core/module_protocols/core_exec.py
@@ -1,0 +1,117 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
+from subprocess import PIPE, Popen
+import copy
+import re
+import datetime
+from core.utility import reverse_and_regex_condition
+from core.utility import process_conditions
+from core.utility import get_dependent_results_from_database
+from core.utility import replace_dependent_values
+
+
+def response_conditions_matched(sub_step, response):
+    conditions = sub_step['response']['conditions']
+    condition_type = sub_step['response']['condition_type']
+    condition_results = {}
+    if sub_step['method'] == 'os_command':
+        if response:
+            response = response.decode()
+            for condition in conditions:
+                if condition == 'END_DATE':
+                    try:
+                        operation = conditions[condition]['regex'].split()[0]
+                        seconds = float(conditions[condition]['regex'].split()[1])
+                        if operation in [">", "<", ">=", "<=", "=="]:
+                            response = re.findall("expires `.*UTC'", response)[0][9:]
+                            end_date = datetime.datetime.strptime(response[:-1], '%Y-%m-%d %H:%M:%S %Z').timestamp()
+                            now_date = datetime.datetime.now().timestamp()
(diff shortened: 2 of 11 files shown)


config.py
```diff
@@ -118,7 +118,7 @@ def nettacker_user_application_config():
         "scan_ip_range": False,
         "scan_subdomains": False,
         "skip_service_discovery": False,
-        "thread_per_host": 100,
+        "thread_per_host": 1024,
```
Ali-Razmjoo, line 121: this is on purpose

modules/scan/port.yaml
```diff
@@ -1027,6 +1027,9 @@ payloads:
         response:
           condition_type: or
           conditions:
+            exec:
```
Ali-Razmjoo, line 1030: not sure if there was a better hack.

modules/scan/soon_to_expire_certificate.yaml
```diff
@@ -0,0 +1,48 @@
...
+                - 684 # Corba IIOP SSL
+                - 695 # IEEE-MMS-SSL
+                - 853 # DNS over TLS
+                - 902 # VMWare Auth Daemon
+                - 989 # FTPS data
+        response:
+          condition_type: or
+          conditions:
+            END_DATE:
+              regex: < 2592000 # seconds (30days)
```
Ali-Razmjoo, line 47: it will grab expired one too, naturally endate < 2592000 and also 0 < 2592000, expired one will be gathered as in "untrusted_certificate_scan" too.

modules/scan/certificate_information.yaml
```diff
@@ -0,0 +1,50 @@
+# echo | openssl s_client -showcerts -connect expired.badssl.com:443
+info:
+  id: untrusted_certificate_scan
```
securestep9, line 3: best to name all ssl-related modules with **ssl_** prefix for a consistent naming convention, e.g.
certificate_information->ssl_certificate_information 
weak_tls_ssl_cipher_suites -> ssl_weak_tls_cipher_suites , insecure_tls_ssl_cipher_suites -> ssl_insecure_tls_cipher_suites, soon_to_expire_certificate -> ssl_soon_to_expire_certificate , untrusted_certificate  -> ssl_untrusted_certificate

core/module_protocols/core_exec.py
```diff
@@ -0,0 +1,116 @@
...
+            if condition_type == 'and':
+                return condition_results if len(condition_results) == len(conditions) else []
+            if condition_type == 'or':
+                return condition_results if condition_results else []
+    return []
+
+
+class NettackerExecute:
+    def os_command(command):
+        p = Popen(command, shell=True, stdout=PIPE, stderr=PIPE)
```
securestep9, line 55: command injection vulnerability here(!)

modules/scan/port.yaml
securestep9: we should not be using 'exec' to call external tools like openssl. There are  **ssl** **cryptography** and **pyopenssl**  libraries available built-into Python which can run natively

core/module_protocols/core_exec.py
Ali-Razmjoo: fixed now

modules/scan/port.yaml
Ali-Razmjoo: Unfortunately pyopenssl seems to not support all the functionality I needed.

PR #610, opened 2022-10-05
push new image to dockerhub from master

Ali-Razmjoo:
auto update docker image


Ali-Razmjoo:
@securestep9 I am pushing the docker towards https://hub.docker.com/r/alirazmjoo/owaspnettacker but now I noticed there is a repo from OWASP, https://hub.docker.com/r/owasp/nettacker would you please update the secrets and image path, let's make automatic push from master

PR #559, opened 2022-04-15
✨Mysql brute force module protocol

Mrinank-Bhowmick:
# Motivation and Context
## Changes proposed in this pull request
* Added Mysql brute module protocol , taken reference from [ftp_brute](https://github.com/OWASP/Nettacker/blob/master/modules/brute/ftp.yaml) .
* Changed Dockerfile to add a line to install [mysql-connector-python](https://pypi.org/project/mysql-connector-python/) library.  
  -  here it is placed separately as it will show error if placed in requirements.txt

### Reviewers
@Ali-Razmjoo
@securestep9

#### Checklist
- [x] I have followed the [Contributor Guidelines](https://github.com/OWASP/Nettacker/wiki/Developers#contribution-guidelines).
- [x] The code has been thoroughly tested in my local development environment with flake8 and pylint.
- [x] The code is Python 3 compatible.
- [x] The code follows the PEP8 styling guidelines with 4 spaces indentation.
- [x] This Pull Request relates to only one issue or only one feature
- [ ] I have referenced the corresponding issue number in my commit message
- [x] I have added the relevant documentation.
- [x] My branch is up-to-date with the Upstream master branch.

#### Your development environment
- OS: `Docker: Debian GNU/Linux 11 (bullseye)`
- OS Version: `11`
- Python Version: `3.9`

PR #621, opened 2022-11-07
Added mysql brute force

Mrinank-Bhowmick:
After solving some merging issues and problems #559 

@Ali-Razmjoo


Mrinank-Bhowmick:
@Ali-Razmjoo Can you please check why the synk security workflow is stopping

PR #748, opened 2023-10-10
Update Docker configuration files

arkid15r:
#### Checklist
- [x] The code is Python 3 compatible.

Update Docker configuration files to minimize the layers amount and make the code more readable.

upd:
The initial heredocs approach was replaced with RUN commands chained with `&& \` due to the unsatisfied buildkit requirements. I added missing `apt-utils` package to resolve `debconf: delaying package configuration, since apt-utils is not installed` and set  DEBIAN_FRONTEND to noninteractive. Also added a couple of cleanup instructions to minimize the image size.


securestep9:
This is not backwardly-compatible with Docker v20.10 as it fails with error:
`Error response from daemon: dockerfile parse error line 9: unknown instruction: MKDIR`

Screenshot 2023-10-23 at 01 17 32

It works in Docker v23.0, however the saving on the image provided by this PR is minimal, so I am not going to merge it just yet

arkid15r:
> It works in Docker v23.0, however the saving on the image provided by this PR is minimal, so I am not going to merge it just yet

Well, that makes sense. I'm surprised it worked at all as heredocs requires buiildkit/buildx. I've changed that part and added a sort of smoke tests for multiple docker engine versions -- 24.0, 23.0, and 20.10 (I guess that's what you were doing manually). It just checks whether the image is buildable. This job is required to be successful for running `test` job now.

PTAL when you have a chance.

Dockerfile
```diff
@@ -1,12 +1,23 @@
...
+
+RUN mkdir -p .data/results && \
+    apt-get update && \
+    apt-get install -y $(cat requirements-apt-get.txt) && \
+    pip3 install --upgrade pip && \
+    pip3 install -r requirements.txt && \
+    pip3 install -r requirements-dev.txt && \
+    apt-get clean && \
+    rm -rf /root/.cache/* && \
+    rm -rf /var/lib/apt/lists/*
```
Ali-Razmjoo, line 17: I am not sure if that's a good idea, it makes it harder to debug for each step if fails, and more time consuming to repeat the process.

arkid15r:
Suppressed by https://github.com/OWASP/Nettacker/pull/894

PR #877, opened 2024-08-25
Added Scan Compare feature

Captain-T2004:
#### Checklist
- [x] The code is Python 3 compatible.

This PR adds a new scan compare feature to Nettacker that aims to make differentiating the results of two scans much easier.

- OS: `Kali GNU/Linux Rolling x86_64 `
- OS Version: `6.8.11-amd64`
- Python Version: `3.11.9`


arkid15r (commented):
Could you take a look at `Uncontrolled data used in path expression` warnings and also a couple of my comments when you get a chance?

nettacker/core/graph.py
```diff
@@ -77,6 +98,20 @@ def build_text_table(events):
...
+    _table = texttable.Texttable()
+    table_headers = list(results.keys())
+    _table.add_rows([table_headers])
+    _table.add_rows(
+        [
+            table_headers,
+            [results[col] for col in table_headers],
+        ]
+    )
+    _table.set_cols_width([len(i) for i in table_headers])
```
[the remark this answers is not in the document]
Captain-T2004: @arkid15r, I have made these changes in a new commit

nettacker/api/engine.py
```diff
@@ -212,6 +213,43 @@ def new_scan():
...
+        if result:
+            return jsonify(
+                structure(
+                    status="success",
+                    msg="scan_comparison_completed",
+                )
+            ), 200
+        return jsonify(structure(status="error", msg="Scan ID not found")), 404
+    except (FileNotFoundError, PermissionError, IOError):
+        return jsonify(structure(status="error", msg="Invalid file path")), 500
```
arkid15r, line 250:
```suggestion
        return jsonify(structure(status="error", msg="Invalid file path")), 400
```

nettacker/core/graph.py
```diff
@@ -168,3 +205,89 @@ def create_report(options, scan_id):
...
+    if not fullpath.startswith(base_path):
+        raise PermissionError
+
+    if (len(fullpath) >= 5 and fullpath[-5:] == ".html") or (
+        len(fullpath) >= 4 and fullpath[-4:] == ".htm"
+    ):
+        html_report = build_compare_report(compare_results)
+        with open(fullpath, "w", encoding="utf-8") as compare_report:
+            compare_report.write(html_report + "\n")
+            compare_report.close()
```
arkid15r, line 274:
```suggestion
```

nettacker/core/graph.py
```diff
@@ -168,3 +205,89 @@ def create_report(options, scan_id):
...
+    elif len(fullpath) >= 5 and fullpath[-5:] == ".json":
+        with open(fullpath, "w", encoding="utf-8") as compare_report:
+            compare_report.write(str(json.dumps(compare_results)) + "\n")
+            compare_report.close()
```
arkid15r, line 278:
```suggestion
```

nettacker/core/graph.py
```diff
@@ -168,3 +205,89 @@ def create_report(options, scan_id):
...
+    elif len(fullpath) >= 5 and fullpath[-4:] == ".csv":
+        keys = compare_results.keys()
+        with open(fullpath, "a") as csvfile:
+            writer = csv.DictWriter(csvfile, fieldnames=keys)
+            if csvfile.tell() == 0:
+                writer.writeheader()
+            writer.writerow(compare_results)
+            csvfile.close()
```
arkid15r, line 286:
```suggestion
```

nettacker/core/utils/common.py
```diff
@@ -350,3 +350,26 @@ def sort_dictionary(dictionary):
...
+        sanitized_path
+    """
+    allowed_pattern = r"^[a-zA-Z0-9_-]+(\.[a-zA-Z0-9_-]+)?$"
+
+    components = re.split(r"[/\\]", path)
+
+    sanitized_components = []
+    for component in components:
+        if re.match(allowed_pattern, component):
+            sanitized_components.append(component)
```
arkid15r, line 371:
```suggestion
sanitized_components = [component for component in components if re.match(allowed_pattern, component)]
```

arkid15r (commented):
Please consider the suggestions below:

nettacker/config.py
```diff
@@ -102,6 +102,7 @@ class PathConfig:
     tmp_dir = CWD / ".data/tmp"
     web_static_dir = PACKAGE_PATH / "web/static"
     user_agents_file = PACKAGE_PATH / "lib/payloads/User-Agents/web_browsers_user_agents.txt"
+    compare_results_base_path = CWD / "results"
```
securestep9, line 105: @Captain-T2004 this is incorrect - compare_results_base_path should be the same as results_dir (see line 101) and as a results I get error
Screenshot 2024-09-13 at 23 16 29

nettacker/config.py
```diff
@@ -145,6 +146,11 @@ class DefaultSettings(ConfigBase):
     usernames_list = None
     verbose_event = False
     verbose_mode = False
+    scan_compare_id = None
+    compare_report_path_filename = "/results_{date_time}_{random_chars}.json".format(
```
securestep9, line 150: @Captain-T2004  shouldn't this be {results_dir} instead of "/results"...?

nettacker/locale/en.yaml
```diff
@@ -117,4 +117,9 @@ username_list: username(s) list, separate with ","
 verbose_mode: verbose mode level (0-5) (default 0)
 wrong_hardware_usage: "You must select one of these profiles for hardware usage. (low, normal, high, maximum)"
 invalid_scan_id: your scan id is not valid!
-
+compare_scans: compare current scan to old scans using the unique scan_id
```
securestep9, line 120: @Captain-T2004 can you add these messages to another language file e.g. Hindi hi.yaml?

nettacker/config.py
Captain-T2004: @securestep9, actually I added the results directory path as a PathConfig object variable. And also here only the name of the file is being generated instead of the complete path. Inside my create_compare_report method in graph.py, I have included the basepath from an PathConfig object and then created a complete path from user input + basepath in PathConfig. This is why I removed {results_dir} from here. The error I was facing was that each time a comparison is ran through API the default file name that was generated here was being rewritten as a new filename was not generated each time. In order to generate a new file name each time I have added some helper functions in /utils/common.py and along with the base filepath from PathConfig I can create a complete path inside my create_compare_report method. If you think this is incorrect do let me know I will reverse it.

nettacker/locale/en.yaml
Captain-T2004: Yeah, sure

nettacker/config.py
Captain-T2004: Yes, I will change that back to results_dir, I actually made a separate directory called /results in my Nettacker directory but it was not commited as it was an empty dir. I did this because after adding the path sanitization the user can only save files in the allowed results path with a modifiable file name. This way I though it would be better to find the results fairly easily. I will revert it back but do you think I should alter the path sanitization to include more paths? not just the allowed results_dir path.

securestep9: @Captain-T2004  Please revert back - There is no need to have additional 'results' folder - Nettacker already has one official results folder in the '.data/' for all the results. In order to make comparison results easier you might consider using a file name like this 'results_compare_{scan_id}' - this way it will be easy to find.

nettacker/config.py
securestep9: @Captain-T2004 Please use  {results_dir}  - this is the only folder which shold be for Nettacker results. The problem with file name conflict / files being overwritten can be solved by adding strings to the file name e.g. 'report_compare_{scan_id}' as scan_is is already random it actually makes sense to use scan_id as part of the results filename to identify the results belonging to a specific scan

securestep9 (commented):
please check the latest comments

Captain-T2004:
@securestep9, I have reverted the results back to results_id. Please review.

securestep9 (approved):

PR #913, opened 2024-09-21
ScanID added to CLI and HTML report for easy identification of results

securestep9:
ScanID output added to the CLI scan start up and completion messaging and the HTML report footer for the easy identification of the results belonging to a specific scan and for scan comparison purposes


arkid15r (approved):
LGTM

// generally string concatenation via `+` is costly in Python

PR #947, opened 2024-10-21
Remove `numpy` dependency

arkid15r:
Stacked on #946 

Remove `numpy` dependency.


securestep9:
Will try to test this on a large number of targets

securestep9:
@arkid15r  There is a conflict to be resolved in nettacker/core/utils/common.py

arkid15r (commented):
PTAL

securestep9 (approved):

PR #982, opened 2024-12-19
Added sanitization to report_path_filename and enhanced authenticatio…

winters0x64:
## Proposed change

Set the authentication cookie's SameSite attribute to Lax by default to mitigate CSRF attacks, particularly during the grace period of Chrome's automatic Lax implementation.

Added input sanitization for the ``report_path_filename`` parameter in the /new/scan endpoint to prevent attempts to overwrite system files.

## Type of change

- [ ] Dependency upgrade

## Checklist


nettacker/api/engine.py
```diff
@@ -191,6 +193,37 @@ def index():
...
+
+    Returns:
+        the sanitized report path filename
+    """
+    filename = secure_filename(os.path.basename(report_path_filename))
+    if not filename:
+        return False
+    if "." in filename:
+        if (
+            filename.endswith(".html")
```
securestep9, line 211: Repetitive calls to filename.endswith() - reduces maintainability and extensibility. Suggested improvement: 

```

# Define a list or tuple of valid extensions
VALID_EXTENSIONS = (".html", ".htm", ".txt", ".json", ".csv")

# Check if the filename has a dot and ends with a valid extension
if "." in filename and filename.endswith(VALID_EXTENSIONS):


```

securestep9 (changes requested):

nettacker/api/engine.py
winters0x64: Hi @securestep9  I've resolved the issue by implementing a fix as you suggested, Now i'm checking if there is a '.' in the filename if there is one then we look if the extension of that file is valid, I'm also accepting filenames with no extensions also because /core/graph.py has a condition which handles filenames which don't have any extensions.

securestep9 (approved):

securestep9:
@winters0x64 pre-commit checks failed - please check/review

winters0x64:
@securestep9 , Thanks for letting me now, will fix this ASAP.

Issue #555, opened 2022-04-12
MySQL not working


umardraz:
Hi I am trying to configure Nettacker with mysql database, but it is not working. here is my config

```
def nettacker_database_config():
    return {
        "DB": "mysql",
        "DATABASE": "nettacker",
        "USERNAME": "owasp",
        "PASSWORD": "mypasswd",
        "HOST": "localhost",
        "PORT": "3306"
    }
```

And here is the error that appear whenever submit new scan
```

Process Process-1:1:
Traceback (most recent call last):
  File "/usr/lib/python3.8/multiprocessing/process.py", line 315, in _bootstrap
    self.run()
  File "/usr/lib/python3.8/multiprocessing/process.py", line 108, in run
    self._target(*self._args, **self._kwargs)
  File "/root/Nettacker/core/scan_targers.py", line 114, in start_scan_processes
    options.targets = expand_targets(options, scan_unique_id)
  File "/root/Nettacker/core/targets.py", line 97, in expand_targets
    multi_processor(
  File "/root/Nettacker/core/scan_targers.py", line 72, in multi_processor
    remove_old_logs(
  File "/root/Nettacker/database/db.py", line 130, in remove_old_logs
    session.query(HostsLog).filter(
AttributeError: 'bool' object has no attribute 'query'

```
Note: the mysql database is empty there is no table in it

securestep9:
fixed in #1001

PR #997, opened 2025-02-02
Fixed README with installation instructions for Docker and fixed MySQL errors due to SQLite default

pUrGe12:
## Proposed change

1. The issue with the current documentation is that it is not giving a complete explanation for how to install nettacker properly and run it. I have updated the README with the following changes:

- Code install docker-compose and docker (the user might face trouble with this if doing this for the first time, like I was) 
- Changed the container name from `nettacker-nettacker-1` to `nettacker` in the second command.
This was because when I tried running it locally on a new machine, I got the following error:
`Error response from daemon: No such container: nettacker-nettacker-1`
Hence, the change.

2. Fixed MySQL errors fixing #999 and #555 

1. Corrected MySQL server name using pymysql.
2. Updated docstring for DbConfig class for MySQL related parameters
3. Corrected `db.py` file to include pymysql, change arguments based on engine name and improved `remove_old_logs` function with pymysql required changes.

## Type of change

- [ ] Dependency upgrade

## Checklist

## Additional comments:

1. The user needs to install pymysql using `pip install pymysql`
2. The user needs to ensure their server can handle multiple requests, they can do that using `SET GLOBAL max_connections = 1000; ` 


securestep9:
Hi @pUrGe12 Re: README changes - these are coming with a bigger documentation updates we are working on including re-vamped Dockerfile. Re: mySQL issues - it appears a misconfiguration and we are working on improving the documentation to make it clearer as well

Issue #1004, opened 2025-02-09
admin_scan output is not giving the right results format


pUrGe12:
Currently if we use the following command

```bash
python3 nettacker.py -i <some_ip> -m admin_scan -o result_admin_scan.html
```
It will run the admin scan properly and most of resulting status codes would be either 404 or 403. This is expected and this happens but in the output HTML file, it shows all of the hit results as `detected`. Moreover it doesn't display which directory was being testing in that particular scan. The following image will clarify what I am talking about (ignore IP addr)

![Image](https://github.com/user-attachments/assets/425632ee-0163-41e8-ae84-525190edee51)

We have to manually check by expanding the arrows only to find a 404 or a 403 response. 

Same goes for the table formed after the admin scan,

![Image](https://github.com/user-attachments/assets/37b8977a-0cb5-43b5-958c-bec588c6d08c)

This is not really helpful because it doesn't give the status codes or the directories it hit.

I hope I am not missing something or using this wrong. If so, please let me know.

securestep9:
@pUrGe12  please follow my comments on PR #1006  and use dir.yaml (dir_scan) module as an example to get the target URL in the logs

pUrGe12:
yes will do! Thanks for the help

PR #1006, opened 2025-02-11
added the hit URLs to the output log for admin_scan, keeping other functionality the same

pUrGe12:
## Proposed change

Adding the directories hit and the status codes to the output table shown to the user in case of `admin_scans`. Related to issue #1004. I have not changed the HTML, JSON or CSV outputs yet. 

## Type of change

`admin_scan` doesn't return a 'log' key and hence its not present inside 'results' leading to it only showing `detected` for admin scans which is not very helpful. So, this PR is to just fix that for admin scans.

- [x] Code refactoring without any functionality changes
- [ ] Dependency upgrade

## Checklist

### DIFF
--- a/nettacker/core/utils/common.py
+++ b/nettacker/core/utils/common.py
@@ -30,15 +30,27 @@ def replace_dependent_response(log, response_dependent):
         return log
 
 
-def merge_logs_to_list(result, log_list=[]):
+def merge_logs_to_list(result, log_list=[], module_name=''):
     if isinstance(result, dict):
-        for i in result:
-            if "log" == i:
-                log_list.append(result["log"])
-            else:
-                merge_logs_to_list(result[i], log_list)
-    return list(set(log_list))
+        if "admin_scan" in module_name.strip():
+            for i in result:
+                # Things important in an admin scan
+                if "url" == i.strip() or "status_code" == i.strip():
+                    print(i)
+                    if "url" == i.strip():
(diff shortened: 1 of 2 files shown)


securestep9:
@pUrGe12  This is an incorrect approach architecturally - you do not modify the core to accommodate one module. Nettacker can log the URL using response_dependent['url'] - please see how it is implemented in dir_scan module (**dir.yaml**) and implement the same in **admin.yaml**

nettacker/core/utils/common.py
```diff
@@ -30,15 +30,27 @@ def replace_dependent_response(log, response_dependent):
...
-def merge_logs_to_list(result, log_list=[]):
+def merge_logs_to_list(result, log_list=[], module_name=''):
     if isinstance(result, dict):
-        for i in result:
-            if "log" == i:
-                log_list.append(result["log"])
-            else:
-                merge_logs_to_list(result[i], log_list)
-    return list(set(log_list))
+        if "admin_scan" in module_name.strip():
```
securestep9, line 35: this is breaking the Nettacker modular design as you are hardcoded a module name in the core and modifying for ust one module instead of using the existing module-agnostic functionality

pUrGe12: Ahh, I see. I will fix that

pUrGe12:
Thanks for the review. I shall make the necessary changes.

Issue #1009, opened 2025-02-12
dir_scan only takes the base organisation and not the path on top of it


pUrGe12:
I am using the following command for running a directory scan

```bash
python3 nettacker.py -i "https://abcd.com/test" -m dir_scan -t 100 -o result.html
```

and the output that I get is the enumeration scan against `abcd.com`. 

The way I would do this using gobuster is 

```bash
gobuster dir -u "https://abcd.com/test/" -w "path/to/wordlist.txt" -t 100
```

and it would give hit results building from `/test`.

Is there a similar way to do that for nettacker?

PR #1010, opened 2025-02-13
added argument to take in base path for the URL supplied for directory enumeration

pUrGe12:
## Proposed change

This PR is to fix the issue described in #1009. Earlier running the following command

```sh
python3 nettacker.py -i https://test_website_name.com/test_dir -m dir_scan
```
would result in a directory enumeration starting not from `/test_dir` but from `https://test_website_name.com/` itself. This was not desired because a most of the times paths are stacked.

With the added parser the user can type in the following command to ensure a proper enumeration

```sh
python3 nettacker.py -i https://test_website_name.com -m dir_scan -B "/test_dir"
```
In this case, the names present in the dir_wordlist build on top the `test_dir`.

I have removed the port option from dir_scan because the `http` and `https` schema do not need that. (other than that, I have ensured that I don't mess with any core functionality)

## Type of change

- [x] New core framework functionality
- [ ] Dependency upgrade

## Checklist

### DIFF
--- a/nettacker/config.py
+++ b/nettacker/config.py
@@ -131,6 +131,7 @@ class DefaultSettings(ConfigBase):
     show_all_profiles = False
     show_help_menu = False
     show_version = False
+    url_base_path = None
     skip_service_discovery = False
     socks_proxy = None
     targets = None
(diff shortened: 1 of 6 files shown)


securestep9:
Hi @pUrGe12 this is not the best solution from the design perspective as you are changing the core of Nettacker to inject changes to be used by one module. You have also removed the :{{ports}} in the URL which means you do not fully understand the Nettacker architecture and its ability to use non-standard ports. There are ways to implement the functionality without introducing the an extra command-line option and breaking the core. I will contact you on Slack to explain and guide you to an alternative solution

pUrGe12:
Alright! I'll be glad to learn and do this the right way.

PR #1012, opened 2025-02-16
updated regex for SSH scanning to include more matches

pUrGe12:
## Proposed change

The current regex for SSH port matching is correct for matching banners from Ubuntu, Debian, FreeBSD SSH banners, however a lot of services have SSH banners that are like this

`SSH-2.0-ad34f3baf` 

For example, running the following reveals the same for github.com,
`nc -v github.com 22`

Thus, to match these as well, I have updated the regex. 

## Type of change

- [x] Code refactoring without any functionality changes
- [ ] Dependency upgrade

## Checklist

## Additional info

Future work includes creating NULL probes to capture banners of services that reveal it without any explicit probing (like SSH, FTP) along with the arbitrary probe that is currently defined.

### DIFF
--- a/nettacker/modules/scan/port.yaml
+++ b/nettacker/modules/scan/port.yaml
@@ -1083,7 +1083,7 @@ payloads:
               reverse: false
 
             ssh:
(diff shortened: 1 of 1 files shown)


securestep9 (approved):

securestep9:
Thank you for your contribution @pUrGe12 !

PR #1019, opened 2025-02-20
Added base path for directory enumeration

### LINKED ISSUE #1009: dir_scan only takes the base organisation and not the path on top of it

pUrGe12:
## Proposed change

Closes #1009. Updated the dir_scan and admin_scan modules to take the base directory from the target whenever specified. This allows Nettacker to perform directory enumeration on top of existing paths, as shown in the image below.

usingNettacker

## Type of change

- [x] New core framework functionality
- [ ] Dependency upgrade

## Checklist


securestep9 (approved):
thank you @pUrGe12 ! 🚀

PR #1042, opened 2025-03-11
Add die.py tests

pUrGe12:
## Proposed change

Implementing test cases for die.py. Tested die_success and die_failure.

## Type of change

- [ ] Dependency upgrade

## Checklist


tests/core/test_die.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+class TestDie(TestCase):
+    @patch("sys.stdout", new_callable=StringIO)
+    @patch("sys.exit")
+    def test_die_success(self, mock_exit, mock_stdout):
+        reset_code = TerminalCodes.RESET.value
+        die_success()
+        success_message = mock_stdout.getvalue()
+        assert reset_code in success_message
```
arkid15r: Please use `self.assert...` assertions for consistency.

arkid15r (changes requested):

tests/core/test_die.py
arkid15r: Alternatively, we could migrate to `pytest` syntax as it's not too late yet.

pUrGe12: That will be much cleaner, if you prefer I can work on migrating all tests to a pytest based ones?

arkid15r: Yes, that's a good idea 👍

PR #1051, opened 2025-03-22
Adding config_file_scan 

Manushya-a:
## Proposed change

This PR aims to add a scan module config_file_scan which searches for configuration file in a website and adds a wordlist for the same 

## Type of change

- [ ] Dependency upgrade

## Checklist

- [x] I've run make pre-commit, it didn't generate any changes
- [x] I've run make test, all tests passed locally


pUrGe12:
Hey, while you're at it, I think creating something like an **extension adder** would also make sense? For example, using the '-x' flag and specifying extensions to add onto the words in the wordlist file you're using while doing a scan would be cool and would probably help in config scans.

Manushya-a:
ya makes sense. I will get to it :D

Manushya-a:
Hey, I am thinking of adding this functionality to all modules that use wordlists. So maybe we should tackle that in a separate pr?

securestep9 (approved):

securestep9:
Thank you for your contribution @Manushya-a ! 🚀

PR #1058, opened 2025-04-05
Python version max limit updated in  poetry.lock and pyproject.toml file

Ap1311:
### 🔧 Fix Python Version Constraint in Poetry Lock File

**problem:**
poetry-error

user with pythons latest version 3.13.2 cannot install with poetry this tool due to max limit of python version , defined at poetry.lock and pyproject.toml file

**📋 Proposed Change**
This PR increases the upper limit of the Python version in the poetry.lock and pyproject.toml file to allow compatibility with Python versions beyond 3.13.0, such as 3.13.2, which previously caused installation errors due to overly strict version constraints.

📎 Context
Previously, the project specified a maximum Python version like:

`python = ">=3.7,<3.13"`

This caused issues when users attempted to install or run the project with Python 3.13.x, receiving errors such as:

`Current Python version (3.13.2) is not allowed by the project (^3.9, <3.13). Please change python executable via the "env use" command. `

allowing users with Python 3.13.x to install and run the project without version lock conflicts.

**🛠 Type of Change**

- Dependency upgrade
-  Bugfix (non-breaking change that fixes version compatibility)

✅ Checklist

- I've followed the [contributing guidelines](https://nettacker.readthedocs.io/en/latest/Developers/)
- Updated the poetry.lock and/or pyproject.toml to reflect version changes
- Tested locally with Python 3.13.2
- Ran make test and make pre-commit — all passed

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -1738,7 +1738,7 @@ files = [
 ]
 
 [package.dependencies]
-greenlet = {version = "!=0.4.17", markers = "python_version < \"3.13\" and (platform_machine == \"aarch64\" or platform_machine == \"ppc64le\" or platform_machine == \"x86_64\" or platform_machine == \"amd64\" or platform_machine == \"AMD64\" or platform_machine == \"win32\" or platform_machine == \"WIN32\")"}
+greenlet = {version = "!=0.4.17", markers = "python_version < \"3.13.2\" and (platform_machine == \"aarch64\" or platform_machine == \"ppc64le\" or platform_machine == \"x86_64\" or platform_machine == \"amd64\" or platform_machine == \"AMD64\" or platform_machine == \"win32\" or platform_machine == \"WIN32\")"}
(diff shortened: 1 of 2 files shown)


securestep9:
@Ap1311 the restriction to disallow Python v3.13 is **intentional** and it is due to a **compatibility issue.** 
At this moment Nettacker is not compatible with Python 3.13 due to multiple breaking changes introduced in Python v3.13  and several important dependencies  breaking the functionality of the core and several modules. Artificially changing the max version in` pyproject.toml `&` poetry.lock `will not fix the code incompatibility. The approved version of Python for Nettacker is 3.11, most modules work up to 3.12.6 but any Python version higher than that will cause multiple compatibility issues in Nettacker. 

If you wish to run Nettacker on your system where you have only Python3.13 you can use virtualenv with Python3.11, for example:

`virtualenv --python="/usr/bin/python3.11" venv311`

PR #1066, opened 2025-05-05
security improvements

securestep9:
## Proposed change

Security improvements:
* added security headers to API engine responses
* prevented Server header from exposing version info, now it returns just "API"

## Type of change

- [ ] Dependency upgrade

## Checklist

### DIFF
--- a/nettacker/api/engine.py
+++ b/nettacker/api/engine.py
@@ -43,6 +43,10 @@ from nettacker.database.db import (
 )
 from nettacker.database.models import Report
 
+# Monkey-patching the Server header to avoid exposing the actual version
+from werkzeug.serving import WSGIRequestHandler
+WSGIRequestHandler.version_string = lambda self: "API"
+
 log = logger.get_logger()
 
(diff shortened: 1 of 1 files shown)


nettacker/api/engine.py
```diff
@@ -154,6 +158,17 @@ def access_log(response):
             ).encode()
         )
         log_request.close()
+        # Set security headers
```
arkid15r, line 161: It seems that the headers will be added when API access logs are enabled only. Is it the intended behavior?

arkid15r (commented):
I'm not familiar w/ Flask comparing to Django -- it seems like it should be a separate middleware file or at least a separate view that is not tied to any other logic other than response header setting.

nettacker/api/engine.py
securestep9: well-spotted! Codex stuck the headers in the access-log function, I will get it fixed and split into a separate function for setting the headers

arkid15r (approved):
LGTM

PR #1074, opened 2025-05-28
robust handling of getservbyport() and filtered port detection

pUrGe12:
## Proposed change

The getservbyport() function relies on the `etc/services` file for default service resolution, however for non-default ports this fails and hence, even if response from that port is received, nettacker fails to report it (as it hits an error). 

Example demonstrated:
image

This is port 8443 filtered. This port is not present in /etc/services file and hence nettacker gives,
image

After the changes made here, Nettacker is able to classify the port as filtered as well. (by updating the `service` field)

image

This is using the same timeout value being previously used inside `create_tcp_socket()`.

## Type of change

- [ ] Dependency upgrade

## Checklist

### DIFF
--- a/nettacker/core/lib/socket.py
+++ b/nettacker/core/lib/socket.py
@@ -24,6 +24,9 @@ def create_tcp_socket(host, port, timeout):
         ssl_flag = False
     except ConnectionRefusedError:
         return None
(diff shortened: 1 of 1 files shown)


pUrGe12:
This requires some more thought.

PR #1078, opened 2025-06-06
added regex for elasticsearch

pUrGe12:
## Proposed change

Added regex matches for [elasticsearch](https://en.wikipedia.org/wiki/Elasticsearch). 

image

## Type of change

- [ ] Dependency upgrade

## Checklist


nettacker/modules/scan/port.yaml
```diff
@@ -1042,6 +1042,10 @@ payloads:
                 regex: "HTTPStatus.BAD_REQUEST|HTTP\\/[\\d.]+\\s+[\\d]+|Server: |Content-Length: \\d+|Content-Type: |Access-Control-Request-Headers: |Forwarded: |Proxy-Authorization: |User-Agent: |X-Forwarded-Host: |Content-MD5: |Access-Control-Request-Method: |Accept-Language: "
                 reverse: false
 
+              elasticsearch:
```
arkid15r, line 1043: Could you move it above `ftp`?

arkid15r (commented):
Looks reasonable, it seems we need to do some re-ordering there (but it's a different PR task).

nettacker/modules/scan/port.yaml
pUrGe12: I did that but I don't think its such a good idea, because I think it cascades down the services list in port.yaml and for elasticsearch services, the returned banner can be identified under `http` as well, so we might end up getting `http` instead of `elasticsearch`

Issue #1116, opened 2025-08-09
Fix needed in webui where select all profiles button working is glitched


Davda-James:
### What is the issue?
So selecting `select all profiles` selects all items of `scan methods` instead of selecting all profiles.

### Proof of issue

![Image](https://github.com/user-attachments/assets/ff6cfa61-dcc4-4310-8100-28309e836834)

Issue #781, opened 2023-12-26
Reorganize Profile Section in Web Module for Improved Usability


prbhtkumr:
## Issue Summary:
The current profile section in the web module appears cluttered with multiple protocols and products, making it challenging for users to navigate and select profiles efficiently. This issue proposes reorganizing the profile section to enhance user experience.

## Proposed Changes:

- **Separate Section for Profiles:**
  Introduce a separate section for profiles to declutter the existing profile selection interface.

- **Improved Categorization:**
  Group profiles based on protocols and products to make it more intuitive for users.

#### Screenshot:
![Screenshot_20231226_010406](https://github.com/OWASP/Nettacker/assets/50882504/13d1b55e-e8bf-49ee-91d0-2a6cf0cf9cc7)

## Implementation Plan:

- Identify the relevant HTML templates and JavaScript files responsible for rendering the web interface.
- Introduce modifications to separate the profile section and improve categorization.

## Benefits:

- Enhances user experience by providing a more organized and user-friendly profile selection interface.
- Simplifies the process of selecting profiles, especially when dealing with multiple protocols and products.

prbhtkumr:
I am new to the project and would like to work on this issue. 
As a newcomer, I'd really appreciate mentorship or guidance through the codebase.
Could you please assign me this project? @securestep9

vikas-chaudhary-2802:
@prbhtkumr Claim
i start working on it

prbhtkumr:
> @prbhtkumr Claim i start working on it

Hey @vikas-chaudhary-2802 
Thank you for showing interest in this issue.

Currently, there's an ongoing code refactoring initiative in progress, and I would like to align my efforts with this issue as I'm relatively new to the project and it offers a great learning opportunity.

I hope you understand.

However, there are plenty of other exciting areas within the project where your contributions would be highly valuable. I encourage you to explore other open issues or functionalities that interest you.

If you have any questions or need assistance finding another area to contribute, feel free to ask. I look forward to seeing your contributions to the project!

securestep9:
There is indeed some refactoring going on. As this improvement was proposed by @prbhtkumr  it is only fair for him to work on it in own fork.   For you @vikas-chaudhary-2802  I suggest to have a look at the the pagination feature in crawler results #805

Davda-James:
@securestep9 exactly can you describe what all enhancements we need in this ?

securestep9:
[...]

PR #1117, opened 2025-08-09
Fixed the issue of select all profiles button

Davda-James:
Earlier `select all profiles` was selecting every `scan methods`  instead of profiles, fixed it now.

### File changed while fixing
`nettacker/web/static/index.html`
`nettacker/web/static/js/main.js`

Solves the issue #1116

I was working on the issue #781, and I first observe this glitch so thought it would be better to create a separate issue of this and make PR for this as this issue is a bit off track from issue #781.

## Type of change

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added a "select/deselect all" option for profile checkboxes in the scan form, allowing users to quickly select or clear all profiles at once.

* **Style**
  * Updated the identifier and CSS class for the "select all profiles" checkbox to improve clarity and consistency.

### DIFF
--- a/nettacker/web/static/index.html
+++ b/nettacker/web/static/index.html
@@ -172,9 +172,9 @@
                 <h3>Profiles</h3>
                 <div class="form-group">
                     <div id="profiles" class="checkbox text-justify">
-                        <label><input id="all" type="checkbox" class="check check-all-scans"><a
-                                class="label label-info">Select all profiles</a></label>&nbsp;&nbsp;&nbsp;&nbsp;
-                        {% autoescape off %}{{profile}}{% endautoescape %}
+            <label><input id="all_profiles" type="checkbox" class="check check-all-profiles"><a
+                class="label label-info">Select all profiles</a></label>&nbsp;&nbsp;&nbsp;&nbsp;
+            {% autoescape off %}{{profile}}{% endautoescape %}
                     </div>
                 </div>
                 <h3>Scan Methods</h3>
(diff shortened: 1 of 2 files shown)


nettacker/web/static/index.html
```diff
@@ -172,9 +172,9 @@ <h3>Targets</h3>
                 <h3>Profiles</h3>
                 <div class="form-group">
                     <div id="profiles" class="checkbox text-justify">
-                        <label><input id="all" type="checkbox" class="check check-all-scans"><a
-                                class="label label-info">Select all profiles</a></label>&nbsp;&nbsp;&nbsp;&nbsp;
-                        {% autoescape off %}{{profile}}{% endautoescape %}
+            <label><input id="all_profiles" type="checkbox" class="check check-all-profiles"><a
```
pUrGe12, line 175: Changing the id would probably break this because it doesn't know what "all_profiles" means. You should keep this as "all"

Davda-James: I have searched for that if there has been any usage of getElementById or in any other form of "all", but I did not get that so I used "all_profiles", if you can find some breakpoints, please suggest me, I will fix it according to that.
Thank You

pUrGe12: What's happening is that when we click on the "select all profiles" button, its taking "all_profiles" as a profile name which is invalid. Can you look into that?

Davda-James: Yeah I looked into it and fixed that, can you please check once.

PR #1145, opened 2025-10-05
fix: database lock issues fixed using litequeue

Davda-James:
## Proposed change

- Queue-based Architecture: Database operations are enqueued and processed by a separate writer thread
- Persistent Storage: Uses litequeue for disk-backed job persistence across process restarts
- Fallback Support: Falls back to in-memory queues and synchronous writes when litequeue is unavailable
- WAL Mode: Enabled SQLite WAL mode for improved concurrent access
- CLI Worker: Added standalone database worker process for production deployments

## Type of change
Feat + Fix 

- [x] Dependency upgrade
- [x] Written test for new change 

## Files Changed
```bash
writer.py
db.py
db_worker.py
pyproject.toml
test_writer.py
```
## Note
pyproject.toml has been updated with litequeue (^0.9), hence poetry.lock

## Checklist

### DIFF
--- a/nettacker/database/writer.py
+++ b/nettacker/database/writer.py
@@ -163,7 +163,6 @@ class DBWriter:
                 return None
 
     def _run(self):
-        session = self.Session()
         pending = []
         while not self._stop.is_set():
             try:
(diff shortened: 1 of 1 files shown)


nettacker/cli/db_worker.py
```diff
@@ -0,0 +1,48 @@
...
+                f"processed={processed} total_processed={stats.get('processed')} queue_size={stats.get('queue_size')}"
+            )
+        return
+
+    try:
+        while True:
+            time.sleep(1)
+    except KeyboardInterrupt:
+        writer.stop()
+
```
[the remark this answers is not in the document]
Davda-James: fixed

nettacker/cli/db_worker.py
pUrGe12, line 1: This is not in accordance to the main architecture. You're defining a new argument parsing module which is not necessary imo. Please look into that.

tests/database/test_writer.py
pUrGe12, line 1: We're using pytest for qa automation

tests/database/test_writer.py
```diff
@@ -0,0 +1,115 @@
...
+        writer.enqueue(job)
+
+    # Process all queued jobs
+    processed_count = writer.drain_once(max_iterations=10)
+    assert processed_count == 2
+
+    # Verify the jobs were written to the database
+    import sqlite3
+
+    conn = sqlite3.connect(db_path)
```
pUrGe12, line 103: You really don't want to make actual live connections during testing. I think you should mock the database call if you're doing unit testing.

nettacker/cli/db_worker.py
Davda-James: Okay will look into that

Davda-James: Hey @pUrGe12 can you tell me how exactly we want so that before starting again would get some better idea for implementation.

PR #1151, opened 2025-10-15
Windows compatibility support 

### LINKED ISSUE #933: Refactor the code to make sure os path related logic is OS agnostic
Refactor the code to make the code expecting Unix/Linux paths OS-independent.
Test and ensure that the the entire application with all its functionality (including Scan engine, API and WebUI) works equally fine on both Windows and Unix-like operating systems

codeshazard:
## Proposed change
This PR adds Windows compatibility to OWASP Nettacker by addressing platform-specific issues that previously prevented it from running on Windows systems.

**Changes:**
- Made uvloop dependency conditional using platform markers (doesn't support Windows)
- Added win32 to allowed platforms in platform check
- Implemented cross-platform privilege checking to replace Unix-only os.geteuid()
- Added explicit UTF-8 encoding for file operations
- Replaced hardcoded path separators with pathlib for cross-platform compatibility

**Related:** Replaces #932 with properly signed commits as requested.

## Type of change

## Checklist

Tested manually on windows since make commands are unix specific


nettacker/core/app.py
```diff
@@ -37,6 +37,24 @@
 log = logger.get_logger()
 
 
+def is_running_with_privileges():
```
pUrGe12: I think it'll be a better idea to keep this function inside common.py and not in the main app file

codeshazard: I agree, thanks for the review!

codeshazard:
@arkid15r @securestep9 , Just following up on this PR. Happy to address any feedback or make additional changes if needed. Thanks!

codeshazard:
Hi @arkid15r  @securestep9, Sorry for the delay, I've regenerated the poetry.lock file to fix the dependency sync issue. Could you please trigger the CI/CD to verify if the changes work correctly? Thanks!

codeshazard:
Hi @arkid15r @securestep9  Just checking in on this PR. It's been a couple of weeks - could you please trigger the CI/CD pipeline or let me know if there's anything I should address?

codeshazard:
@securestep9 @arkid15r Thanks for triggering the pipeline.  All CI/CD jobs have passed successfully.  Please let me know if anything needs to be improved .

securestep9:
@coderabbit review

securestep9:
@codeshazard a few fixes on your PR are still needed - can you please address?

- in `pyproject.toml `a `sys_platform` check is needed for `impacket` (similar to `uvloop` as `impacket` is not compatible with Windows)
- in `nettacker/core/lib/http.py` uvloop import and usage needs to be skipped if the os is Windows
- Path handling improvements still missing in some places

codeshazard:
@securestep9 Sorry for the late reply! I'll work on it asap!

Chirag6722:
I was testing Nettacker on Windows 11 / Python 3.12.10 and ran into the uvloop install blocker this PR fixes, so I applied the change locally to get further. Two data points from that run that might be useful here.

**1. impacket 0.11.0 appears to work on Windows.**

@securestep9 suggested adding a `sys_platform` marker for `impacket` alongside `uvloop`. I couldn't reproduce an incompatibility — on Windows 11 / Python 3.12.10 it installs, and the code path Nettacker actually uses works:

```
>>> from impacket.smbconnection import SMBConnection
smbconnection import OK
>>> from impacket import ntlm; ntlm.NTLMAuthNegotiate()
ntlm OK True
```

`nettacker/core/lib/smb.py:1` is the only consumer (`from impacket.smbconnection import SMBConnection`). If impacket were marked `sys_platform != 'win32'`, that import would fail and the SMB engine would be unavailable on Windows — so the marker may cost working functionality rather than fix anything. Happy to be wrong if there's a specific failure I should be reproducing instead.

**2. The uvloop dependency marker alone doesn't unblock Windows.**

With only the `pyproject.toml` marker applied, the build failure becomes an import failure, because `nettacker/core/lib/http.py` imports uvloop unconditionally at module level:

```python
10:  import uvloop
20:  asyncio.set_event_loop_policy(uvloop.EventLoopPolicy())
```

```
ERROR tests/core/lib/test_user_added_http_headers.py
nettacker/core/lib/http.py:10: in <module>
    import uvloop
E   ModuleNotFoundError: No module named 'uvloop'
```

That matches the second item in @securestep9's 16 Jan review. Flagging it only because it's easy for it to look done once the marker lands — the HTTP engine still can't be imported on Windows until the import itself is guarded.

Not looking to take anything over here, @codeshazard — just passing along what the run showed in case it saves you time.

PR #1155, opened 2025-10-24
Improve robustness: canonicalize & validate targets before scan

Aarush289:
## Proposed change

Introduce a **fast, parallel pre-check** stage that validates targets **before** scanning and **drops invalid/bogus entries early**.

Current behavior attempts socket connects on every port; for hostnames this triggers DNS lookups that can block for a long time on nonsense names. With multiple ports, bogus inputs can consume **minutes** and significant resources.

**This PR:**

* Runs a lightweight, parallel **target validation** pass via `resolve_quick(...)`:

  * Rejects obviously invalid IPv4/IPv6 literals (bad octets, syntax).
  * For hostnames, verifies label syntax and attempts a **short-timeout** A/AAAA resolve.
* **Skips the scan stage** for rejected targets, preventing expensive connect/DNS fallbacks.
* **Canonicalizes** valid targets (e.g., lowercases hostnames, strips trailing dot) and **preserves input order**.
* Adds only a small up-front delay (seconds for large batches) while saving minutes on bad inputs.

**Privacy safeguard (proxy mode):**
If the user requests a proxy (e.g., `-R` or `--socks-proxy`), the pre-check is **disabled** to avoid any **local DNS lookups** that could expose the user’s IP to their ISP/resolver. In proxy mode, targets proceed directly to the scanning path (letting the proxy handle DNS), preserving user anonymity.

**Examples dropped early now:** `290.52.245.54` (invalid IPv4), `google` (no resolvable FQDN), random strings like `dsjfiejb`.

## Type of change

* [ ] New core framework functionality
* [ ] Bugfix (non-breaking change which fixes an issue)
* [ ] Code refactoring without any functionality changes
* [ ] New or existing module/payload change
* [ ] Documentation/localization improvement
* [ ] Test coverage improvement
* [x] Other improvement (best practice, cleanup, optimization, etc)

## Checklist

* [x] I've followed the [[contributing guidelines](https://nettacker.readthedocs.io/en/latest/Developers/)][contributing-guidelines]
* [x] I've run `make pre-commit`, it didn't generate any changes
* [x] I've run `make test`, all tests passed locally

### Additional context

* Adds and uses `resolve_quick(...)` from `app.py` prior to scheduling scans.
* Bounded concurrency (e.g., `min(len(targets), 64)`) to avoid spikes.
* Logs dropped targets with reasons (invalid IP, DNS NXDOMAIN/timeout, bad label).
* Automatically **bypasses pre-check** when `-R/--socks-proxy` is set to prevent local DNS leaks.


nettacker/core/hostcheck.py
```diff
@@ -0,0 +1,163 @@
...
+    seen, uniq = set(), []
+    for c in candidates:
+        if c not in seen:
+            seen.add(c); uniq.append(c)
+    candidates = uniq
+    if not candidates:
+        return False, None
+
+    for _pass_ix, (use_ai_addrconfig, port) in enumerate(((True, None), (False, None))):
+        deadline = time.monotonic() + timeout_sec
```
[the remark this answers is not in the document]
Aarush289: Timeout is intended for per pass .

nettacker/core/hostcheck.py
```diff
@@ -0,0 +1,166 @@
...
+    parts = host.split(".")
+    if len(parts) < 2 and not allow_single_label:
+        return False
+    return all(_LABEL.match(p) for p in parts)
+
+def _system_search_suffixes() -> list[str]:
+    # Only used when host has no dots; mirrors OS resolver search behavior (UNIX).
+    sufs: list[str] = []
+    try:
+        with open("/etc/resolv.conf") as f:
```
securestep9: @Aarush289 using /etc/resolv.conf here is not acceptable. This will not work on Windows and other non-Linux platforms and we have just spent lots of effort making Nettacker platform-independent

securestep9:
Can you also provide justification/reasoning behind this proposed PR. Nettacker mainly operates on Networks (as the name suggests) using IP ranges and subdomain discovery. While theoretically it can be fed whole bunch of unresolvable hostnames as targets our standard recommended way of validating them is to use ping_before_scan option which will filter out non-pingable hosts. I suggest that you re-work your proposal as an **option** e.g. **validate_before_scan**  rather than modification of the core functionality. This way the users will have a choice to turn the validation on or off

Aarush289:
Hi @securestep9 ,

Thank you for the thoughtful review and for pointing out the OS-specific concerns and the risk of changing default behavior. You’re absolutely right—this should be an **opt-in** feature rather than a core change.

A few notes on my original motivation:

* While `ping_before_scan` is useful, ICMP reachability can be an unreliable filter in practice: many networks drop ICMP for policy/privacy, and transient host issues can cause false negatives.
* A **DNS reachability check** (e.g., resolving A/AAAA) can often be a more robust pre-filter for typical Nettacker workflows that start from domain names. (Of course, it’s not infallible—split-horizon DNS, stale cache, or transient resolver issues can still occur—but it generally avoids the ICMP-blocked case and is a more reliable option .)

As Per your guidance, I’ll rework the PR as an **optional feature** rather than modifying core flow. Concretely, I plan to:

1. Add a new flag (name as you suggested): `--validate_before_scan`
2. Make the implementation **cross-platform** (no OS-specific calls), using stdlib wherever possible and clean fallbacks.

Thank you again for the clear direction and for flagging the Windows issue—totally agree it shouldn’t be OS-specific. I’ll incorporate your feedback and resubmit as an opt-in feature with proper documentation.

Best,
Aarush

nettacker/core/app.py
```diff
@@ -289,7 +289,66 @@ def scan_target(
...
+                    canon_by_index[idx] = canon
+                else:
+                    log.info(f"Invalid target -> dropping: {orig_target}")
+
+        # Keep order, drop Nones
+        filtered = [c for c in canon_by_index if c is not None]
+
+        if dedupe:
+            seen, unique = set(), []
+            for c in filtered:
```
pUrGe12: Consider better names, if iterator is not being used elsewhere then just use an "_"

nettacker/core/app.py
```diff
@@ -289,7 +289,66 @@ def scan_target(
...
+            raise TypeError(f"`targets` must be iterable, got {type(targets).__name__}")
+
+        if not targets:
+            return []
+
+        if max_workers is None:
+            max_workers = min(len(targets), 10)  # cap threads
+
+        # Preserve order
+        canon_by_index = [None] * len(targets)
```
pUrGe12: The names can be clearer

nettacker/core/app.py
```diff
@@ -289,7 +289,66 @@ def scan_target(
...
+            ok, canon = resolve_quick(t, timeout_sec=timeout_per_target)
+            return idx, t, (canon if ok and canon else None)
+
+        with ThreadPoolExecutor(max_workers=max_workers) as ex:
+            futures = [ex.submit(_task, i, t) for i, t in enumerate(targets)]
+            for fut in as_completed(futures):
+                try:
+                    idx, orig_target, canon = fut.result()
+                except (OSError, socket.gaierror) as exc:
+                    log.debug(f"Invalid target (resolver error): {exc!s}")
```
pUrGe12: The logger doesn't have a `.debug()` method. Change this to `.error()`.

nettacker/core/hostcheck.py
```diff
@@ -0,0 +1,118 @@
...
+log = logger.get_logger()
+
+_LABEL = re.compile(r"^(?!-)[A-Za-z0-9-]{1,63}(?<!-)$")
+
+def is_ip_literal(name: str) -> bool:
+    """Return True if name is a valid IPv4 or IPv6 address literal."""
+    try:
+        socket.inet_pton(socket.AF_INET, name)
+        return True
+    except OSError:
```
pUrGe12: exceptions are slower than if-else statements. You're better off with precompiling an IPv4 regex (IPv6 would be too complex) at least and not relying on a try-except block as part of the main execution flow.

nettacker/core/hostcheck.py
```diff
@@ -0,0 +1,118 @@
...
+    Args:
+        host: Hostname to validate.
+        allow_single_label: If True, accept single-label names (e.g., "localhost").
+    
+    Returns:
+        True if the hostname is syntactically valid.
+    """
+    if host.endswith("."):
+        host = host[:-1]
+    if len(host) > 253:
```
pUrGe12, line 61: Do add a comment here to specify this comes from RFC1123 which specifies the number of characters to be 250 at max (without dots) and 253 with dots.

nettacker/locale/ar.yaml
pUrGe12, line 1: You don't have to translate for all. Just `en.yaml` is enough, we'll leave the translations to the native speakers.

nettacker/core/hostcheck.py
```diff
@@ -0,0 +1,118 @@
...
+
+
+def _gai_once(name: str, use_ai_addrconfig: bool, port):
+    flags = getattr(socket, "AI_ADDRCONFIG", 0) if use_ai_addrconfig else 0
+    return socket.getaddrinfo(
+        name, port, socket.AF_UNSPEC, socket.SOCK_STREAM, 0, flags
+    )
+
+def _clean_host(s: str) -> str:
+    # remove surrounding quotes and whitespace, lone commas, repeated dots
```
pUrGe12: Are you handling repeated dots?

PR #1177, opened 2025-12-10
Fix: Python 3.12 SSL compatibility - replace ssl.wrap_socket

deekshithaby:
## Proposed change

This PR fixes Python 3.12 compatibility issues in the Nettacker project. Python 3.12 removed the deprecated `ssl.wrap_socket()` function, causing 2 tests to fail with `AttributeError`. 

This PR updates both the source code and tests to use the modern `ssl.SSLContext().wrap_socket()` approach, which is the recommended method for SSL/TLS connections in Python 3.10+.

**Changes:**
- Replaced `ssl.wrap_socket()` with `ssl.SSLContext(ssl.PROTOCOL_TLS_CLIENT).wrap_socket()` in `nettacker/core/lib/ssl.py` and `nettacker/core/lib/socket.py`
- Added `server_hostname` parameter for proper SSL/TLS certificate validation
- Updated test mocks in `tests/core/lib/test_ssl.py` and `tests/core/lib/test_socket.py` to patch the correct SSL method
- All 214 tests now pass on Python 3.12

**Testing:**
- Ran full test suite on Python 3.12.3
- Before: 212 passed, 2 failed
- After: 214 passed, 0 failed

## Type of change

- [x] Bugfix (non-breaking change which fixes an issue)

## Checklist

- [x] I've followed the contributing guidelines
- [x] I've run `make test`, all tests passed locally

Fixes #1190

PR #1180, opened 2025-12-12
tests: add compatibility shim for ssl.wrap_socket on Python 3.12

deekshithaby:
## Problem
Python 3.12 removed `ssl.wrap_socket`, which causes test failures in `test_ssl.py` and `test_socket.py` even though Nettacker's production code correctly uses `SSLContext.wrap_socket`.

This prevents contributors from running tests successfully on Python 3.12+, making it harder to test and review PRs like #1173.

## Solution
This PR adds a small test-only compatibility shim through pytest's configuration system (`tests/conftest.py`). The shim reintroduces `ssl.wrap_socket` by delegating to an `SSLContext` created via `create_default_context()`.

## Changes Made
- **Added `tests/conftest.py`**: Provides SSL compatibility shim for Python 3.12
- **Fixed `tests/common.py`**: Corrected import to use `tests.conftest` instead of missing root `conftest`
- **Defined missing variables**: Added `nettacker_dir` and `tests_dir` that `tests/common.py` expects

## Testing
✅ All 214 tests pass on Python 3.12.3  
✅ No import errors  
✅ SSL-related tests work correctly  

## Benefits
- ✅ Does NOT modify runtime/production code
- ✅ Only affects tests
- ✅ Unblocks test execution on Python 3.12+
- ✅ Improves contributor experience on newer Python versions
- ✅ Allows proper testing of PRs like #1173

## Note to Maintainers
If you prefer updating the tests to mock `SSLContext.wrap_socket` directly instead of using this shim, I'm happy to adjust the approach. This solution prioritizes minimal changes and backwards compatibility.

## Related
- Helps verify #1173 works correctly on Python 3.12
-

### DIFF
--- a/tests/common.py
+++ b/tests/common.py
@@ -1,7 +1,7 @@
 import unittest
 from pathlib import Path
 
-from conftest import nettacker_dir, tests_dir
+from tests.conftest import nettacker_dir, tests_dir
 
 
 class TestCase(unittest.TestCase):
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,9 +1,18 @@
-import sys
-from os.path import abspath, dirname, join
+import ssl
+from pathlib import Path
 
-project_root = dirname(dirname(__file__))
-nettacker_dir = abspath(join(project_root, "nettacker"))
-tests_dir = abspath(join(project_root, "tests"))
+# Define directory paths for tests/common.py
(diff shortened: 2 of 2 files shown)


deekshithaby:
After reflection, PR #1177 is the better approach since it fixes 
the source code directly rather than working around the issue in tests. 
Closing this in favor of that. Thanks!

PR #1184, opened 2025-12-14
Fix SSH brute-force success detection and error handling

berardifra:
### Summary
This PR improves the SSH brute-force module reliability by fixing authentication success detection and making error handling consistent with Nettacker’s engine expectations.

### What was changed
- Updated SSH authentication logic to correctly detect successful logins
- Ensured the module returns a non-empty dictionary only on real authentication success
- Handled authentication, SSH protocol, and network errors safely without crashing scans
- Kept behavior consistent with existing Nettacker modules

### How it was tested
- Local SSH server on 127.0.0.1 (OpenSSH)
- Valid credentials → successful detection in `ssh_brute`
- Invalid credentials → no false positives
- Verified module execution through Nettacker CLI
- Python syntax validation with `py_compile`

### Impact
- Prevents false negatives and unstable behavior in `ssh_brute`
- Improves scan reliability without changing module interface
- No breaking changes

### Checklist
- [x] Code follows project structure
- [x] No YAML or interface changes required
- [x] Tested with real SSH service

### DIFF
--- a/nettacker/core/lib/ssh.py
+++ b/nettacker/core/lib/ssh.py
@@ -1,12 +1,10 @@
 import logging
-
-from paramiko import SSHClient, AutoAddPolicy
+import socket
+from paramiko import SSHClient, AutoAddPolicy, AuthenticationException, SSHException
 from paramiko.auth_strategy import NoneAuth, Password
-
 from nettacker.core.lib.base import BaseEngine, BaseLibrary
 
 logging.getLogger("paramiko.transport").disabled = True
-
 
 class SshLibrary(BaseLibrary):
     client = SSHClient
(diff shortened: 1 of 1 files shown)


nettacker/core/lib/ssh.py
```diff
@@ -1,41 +1,57 @@
...
-            "port": port,
-            "username": username,
-            "password": password,
-        }
 
+        except AuthenticationException:
+            return {}
+        except SSHException:
+            return {}
+        except (socket.timeout, socket.error, ConnectionRefusedError, OSError):
```
pUrGe12, line 46: You don't need to use both OSError and socket.error. Python3+ they are basically aliases. Also, you really aren't doing anything special by handling them separately.

nettacker/core/lib/ssh.py
```diff
@@ -1,41 +1,57 @@
...
-
 class SshLibrary(BaseLibrary):
     client = SSHClient
 
     def brute_force(self, *args, **kwargs):
         host = kwargs["host"]
         port = kwargs["port"]
         username = kwargs["username"]
-        password = kwargs["password"]
+        password = kwargs.get("password")
```
pUrGe12, line 16: don't change what's not broken

nettacker/core/lib/ssh.py
```diff
@@ -1,41 +1,57 @@
 import logging
-
-from paramiko import SSHClient, AutoAddPolicy
+import socket
```
pUrGe12, line 2: you "probably" don't need this here... look at the other comment below

pUrGe12 (commented):
You should remove the bloat code and please run pre-commit for ruff to handle the styling.

PR #1240, opened 2026-01-23
Add FortiWeb Authentication Bypass Detection for CVE-2025-64446

Aarush289:
## Proposed change

This PR adds a new vulnerability module to detect **CVE-2025-64446**, a critical authentication bypass vulnerability affecting **Fortinet FortiWeb**.

The vulnerability arises from a relative path traversal issue that allows unauthenticated access to an internal CGI endpoint, enabling attackers to perform administrative actions via crafted API requests. The module sends a carefully constructed HTTP request to the vulnerable endpoint and validates the response using strict status code and content-based matchers to identify affected systems.
## Type of change

## Checklist

Tested against mock server .
Vulnerable FortiWeb images are not publicly available, preventing local deployment.
The module reproduces the same API endpoint, headers, and response patterns described in the official advisories and in other vulnerability tools .
To prevent false negatives in case of multiple scans ( server might give an error if we create user with hardcoded username ( user already exist ) ) , so I have exposed the scan_id to modules and using that as user name .
I noticed that in wp_plugin_cve_2021_39320 , it was using field "status" which doesn't even exist so I updated that too in this PR ( small bug fix )


nettacker/modules/vuln/fortiweb_auth_bypass_cve_2025_64446.yaml
```diff
@@ -0,0 +1,97 @@
...
+                  "history-password5": "",
+                  "history-password6": "",
+                  "history-password7": "",
+                  "history-password8": "",
+                  "history-password9": "",
+                  "force-password-change": "disable",
+                  "force-password-change_val": "0",
+                  "password": "nettacker_pass123"
+              }
+            }
```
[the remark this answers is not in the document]
Aarush289: Warning added

nettacker/modules/vuln/fortiweb_auth_bypass_cve_2025_64446.yaml
```diff
@@ -0,0 +1,97 @@
...
+                  "history-password4": "",
+                  "history-password5": "",
+                  "history-password6": "",
+                  "history-password7": "",
+                  "history-password8": "",
+                  "history-password9": "",
+                  "force-password-change": "disable",
+                  "force-password-change_val": "0",
+                  "password": "nettacker_pass123"
+              }
```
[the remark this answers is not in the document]
Aarush289: I would ask the maintainer if I can  add the feature of generating random ids in these modules ( that would be helpful )

Aarush289:
Hi @securestep9 , 
I had a quick question regarding module capabilities. For cases like this (where a fixed identifier can cause false negatives on re-scan), would you be open to supporting per-scan random variable generation in Nettacker modules (e.g., randomized usernames or IDs)?
This would allow intrusive modules to avoid collisions across runs without adding unsafe cleanup logic, and could also improve reliability for other state-modifying checks.

nettacker/modules/vuln/wp_plugin_cve_2021_38314.yaml
securestep9, line 1: @Aarush289 out of scope change will not be accepted. Please keep your PRs atomic and focused on solving one thing only at a time

nettacker/core/module.py
```diff
@@ -39,6 +39,7 @@ def __init__(
                 self.module_inputs[module_extra_args] = self.module_inputs["modules_extra_args"][
                     module_extra_args
                 ]
+        self.module_inputs["scan_id"] = scan_id
```
securestep9, line 42: This is not related to Fortiweb YAML. Please avoid modifying Nettacker core or CI/CD

securestep9 (changes requested):

nettacker/modules/vuln/wp_plugin_cve_2021_38314.yaml
Aarush289: @securestep9 , I will create a separate PR for the bug fix , but please let me know if any other changes are needed  or specifically about the exposure of scan id to the modules because that is needed for this module  or if you have any other suggestions regarding generation of a random id for modules !

nettacker/core/module.py
Aarush289: I am using the scan id in the module ,If I use a hard coded id to create the user then it may give false negatives while scanning the target multiple times because this scan creates a user at the target and if we try to create same user again then it might give error leading to false negatives and that's the reason I used scan id as user id

securestep9:

PR #1242, opened 2026-01-24
Fix header matcher key to headers in modules ( Fix #1241)

### LINKED ISSUE #1241: Silent false positives when using "header" instead of "headers" in HTTP response matchers
### **Description**

The HTTP response matcher silently ignores unsupported condition keys.
As a result, modules using `header` (singular) instead of the supported `headers` key can produce **false positives**, since the condition is never evaluated.

This behavior is easy to miss and currently affects multiple existing modules.

---

### **Root Cause**

In `response_conditions_matched()`, only the following condition keys are handled:

* `reason`
* `status_code`
* `content`
* `url`
* `headers`
* `responsetime`

Any other key (for example, `header`) is silently ignored:

```python
for condition in conditions:
    if condition in ["reason", "status_code", "content", "url"]:
        ...
    if condition == "headers":
        ...
```

Ignored conditions do not contribute a failure, so the matcher may still succeed.

---

### **Impact**

* Header checks may never be enforced
* Scans can incorrectly report vulnerabilities
* No warning or error is emitted
* Difficult to debug and easy to misconfigure modules
---

For now, I am updating modules that use the `header` field to `headers` and would appreciate guidance from @securestep9 on a long-term solution, or whether we should formally document the supported condition keys. I’ve also noticed usage of `status` instead of `status_code` in some modules, which I fixed in a previous PR.

Aarush289:
## Proposed change

This PR fixes incorrect usage of the `header` response matcher key in existing modules by replacing it with the supported `headers` key.
This prevents silent false positives caused by header conditions being ignored by the HTTP matcher logic (Fixes #1241).

## Type of change

* [x] Bugfix (non-breaking change which fixes an issue)

## Checklist
* [x] I've run `make pre-commit`, it didn't generate any changes
* [x] I've run `make test`, all tests passed locally

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -2002,14 +2002,14 @@ files = [
 
 [[package]]
 name = "urllib3"
-version = "2.6.0"
+version = "2.6.3"
 description = "HTTP library with thread-safe connection pooling, file post, and more."
 optional = false
 python-versions = ">=3.9"
 groups = ["main"]
 files = [
-    {file = "urllib3-2.6.0-py3-none-any.whl", hash = "sha256:c90f7a39f716c572c4e3e58509581ebd83f9b59cced005b7db7ad2d22b0db99f"},
-    {file = "urllib3-2.6.0.tar.gz", hash = "sha256:cb9bcef5a4b345d5da5d145dc3e30834f58e8018828cbc724d30b4cb7d4d49f1"},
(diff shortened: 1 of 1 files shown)


Aarush289:
Hi @securestep9 , pre-commit checks are getting passed in my local system after pulling the changes  ! I don't know why is it failing here . 
image

securestep9 (approved):

securestep9:
@Aarush289 this was due to a GitHub outage: https://www.githubstatus.com/incidents/xwn6hjps36ty

> Hi @securestep9 , pre-commit checks are getting passed in my local system after pulling the changes ! I don't know why is it failing here . image

securestep9:
Thank you @Aarush289 !

PR #1260, opened 2026-02-03
Improve SSL certificate error handling and logging visibility

### LINKED ISSUE #1179: Improve Error Handling Consistency in Scanning Engine

lvb05:
## Proposed change 
This PR improves SSL certificate error handling and logging visibility in the scanning engine.
It ensures that failures during SSL certificate retrieval do not lead to inconsistent state or misleading results.
Fixes #1179 
Key improvements:

- Preserve `ssl_flag` when certificate retrieval fails
- Prevent incorrect SSL status changes when certificate fetch errors occur
- Improve clarity of SSL-related log messages
- Align SSL error output with the localization system (`locale/en.yaml`)

These changes improve robustness and observability without modifying scan logic or adding new features.

## Type of change

## Checklist


lvb05:
Hey @securestep9 
Whenever possible, please review this!

securestep9:
@lvb05 please review and address comments raised by CodeRabbit. Please also attach a screenshot of Nettacker (with command line you used to launch  Nettacker) showing the error message/crash that you are trying to improve by this PR.

Please also explain this:
"Prevent incorrect SSL status changes when certificate fetch errors occur" - Nettacker does not change 'SSL status'. It only detects situations where SSL certificate has weak ciphers or the certificate is expired - what 'status' are you talking about here?

lvb05:
@securestep9 
The `ssl_certificate_fetch_failed` string was intentionally added only to `en.yaml`.  
For other locales, the key is currently present but empty to avoid duplicating English strings in non-English files.

There are two possible approaches here, and I’d like to confirm the project preference:

1. Keep the key empty in non-English locales and accept blank messages for those locales until proper translations are added.
2. Copy the English fallback string into all locale files to ensure a visible message for non-English users.

For the i18n comment (`ssl_certificate_fetch_failed` in non-English locales), I paused here and asked as this is a policy decision (English-only vs duplicated fallback strings). I’ll apply the preferred approach once confirmed.

Also, about the “SSL status” wording, Nettacker does not change an explicit `SSL_status` field. What I meant was **incorrect evaluation of SSL-related scan conditions** when certificate retrieval fails.

Before this PR:
- If `ssl_flag` was true but certificate fetching failed, `ssl_certificate_scan` returned a partial response without cert-related keys.
- Downstream, `response_conditions_matched` accessed these keys directly, leading to `KeyError` or incorrect rule evaluation.
- This could result in misleading scan outcomes (conditions evaluated incorrectly or scan crashing), even though SSL itself was not weak.

This PR does **not** change how SSL is classified.
It ensures:
- Certificate fetch failures do not crash the scan
- SSL-related conditions are evaluated safely and consistently
- Logs clearly show why certificate analysis failed

I’ll update the PR description to avoid using the term “SSL status” and instead refer to “SSL condition evaluation and error handling”.

PR #1304, opened 2026-02-23
Upgrade API Documentation to be Python 3 Compatible

jess-tech-lab:
## Proposed change

  This PR updates the API document [API.md](/OWASP/Nettacker/blob/master/docs/API.md) to transition all Python examples from Python 2 to Python 3.10+. The current documentation contains deprecated syntax including `print ...`, the `u'` string prefix, and missing byte decoding, which leads to errors in Python 3.

Proposed changes are as follows:

- Byte-to-String Handling: Updated all `r.content` examples to include `.decode('utf-8')` to prevent TypeError in Python 3.
- SSL/TLS Configuration: Added explicit `verify=False` to all example requests to prevent the error `certificate verify failed: self-signed certificate` since by default Python 3 enforces certificate verification and the API uses a self-signed certificate.
- Syntax Change: Changed `print ...` to `print(...)` and removed Python 2 specific unicode prefix `u'`.

## Type of change

- [x] Documentation/localization improvement
- [ ] Dependency upgrade

## Checklist


jess-tech-lab:
@arkid15r and @securestep9 - can you review? Thanks.

PR #1335, opened 2026-03-01
Replace exec() in responsetime condition with safe operator mapping

Prajwal5755:
## Proposed change

This PR replaces the use of 'exec()' in the 'responsetime' condition exvaluation inside 'response_conditions_matched()' function with a safe operator mapping using Python's built-in 'operator' module

Previously, dynamic code execution was used to evaluate comparision operators, which:

* Introduced unnecessary security risk
* Reduced readability
* Made debugger harder

This change:
* Removes dynamic execution
* Uses explicit operator mapping like('operator.eq')
* Validates input format before evaluation

More importantly no functional behavior has been changed.

## Type of change

- [ x ] Code refactoring without any functionality changes
- [ ] Dependency upgrade

## Checklist

- [ x ] I've followed the [contributing guidelines][contributing-guidelines]


Gauravsharma2040:
Hi @Prajwal5755, thanks for working on this this is an important fix 👍

I noticed this PR addresses the same root cause as the issue I recently opened (#1408), which also proposes replacing the exec()-based evaluation in response_conditions_matched() with a safe operator mapping.

To avoid duplicate work and review conflicts, it might be helpful to open an issue before submitting a PR for core engine changes, especially in shared components like http.py. That makes it easier to coordinate fixes and prevents multiple PRs targeting the same code path at the same time.

In this case both approaches are valid, and it may make sense to merge the ideas from both PRs into a single fix (for example, using the operator module while keeping the additional guards) 
Thanks again for the contribution

Prajwal5755:
> Hi @Prajwal5755, thanks for working on this this is an important fix 👍
> 
> I noticed this PR addresses the same root cause as the issue I recently opened (#1408), which also proposes replacing the exec()-based evaluation in response_conditions_matched() with a safe operator mapping.
> 
> To avoid duplicate work and review conflicts, it might be helpful to open an issue before submitting a PR for core engine changes, especially in shared components like http.py. That makes it easier to coordinate fixes and prevents multiple PRs targeting the same code path at the same time.
> 
> In this case both approaches are valid, and it may make sense to merge the ideas from both PRs into a single fix (for example, using the operator module while keeping the additional guards) Thanks again for the contribution

Hi, thanks for pointing this out and for opening issue #1408.

I wasn’t aware of the issue earlier, but I see both approaches are addressing the same root cause. In my PR I’ve also replaced `exec()` with a safer operator-based approach.

I agree it’s good to align solutions and avoid duplication. I’m open to making changes if needed based on maintainer feedback.

Thanks for the suggestion, I’ll make sure to check for existing issues before submitting similar changes going forward.

Gauravsharma2040:
> > Hi @Prajwal5755, thanks for working on this this is an important fix 👍
> > I noticed this PR addresses the same root cause as the issue I recently opened (#1408), which also proposes replacing the exec()-based evaluation in response_conditions_matched() with a safe operator mapping.
> > To avoid duplicate work and review conflicts, it might be helpful to open an issue before submitting a PR for core engine changes, especially in shared components like http.py. That makes it easier to coordinate fixes and prevents multiple PRs targeting the same code path at the same time.
> > In this case both approaches are valid, and it may make sense to merge the ideas from both PRs into a single fix (for example, using the operator module while keeping the additional guards) Thanks again for the contribution
> 
> Hi, thanks for pointing this out and for opening issue #1408.
> 
> I wasn’t aware of the issue earlier, but I see both approaches are addressing the same root cause. In my PR I’ve also replaced `exec()` with a safer operator-based approach.
> 
> I agree it’s good to align solutions and avoid duplication. I’m open to making changes if needed based on maintainer feedback.
> 
> Thanks for the suggestion, I’ll make sure to check for existing issues before submitting similar changes going forward.
Just a small note, since this PR is fixing the same issue discussed in #1408, it might help to link the issue in the PR description (for example with `Fixes #1408` or `Related to #1408`).
That usually makes it easier for reviewers to see the context and keeps the discussion in one place, especially since there are multiple PRs touching the same part of `http.py`.

securestep9:
We cannot accept PRs with unsigned commits and without a corresponding issue opened, linked and assigned. Please check the [Contributor Guidelines](https://github.com/OWASP/Nettacker/blob/master/docs/Developers.md#contribution-guidelines)

Prajwal5755:
Hi @securestep9, this was my very first PR, and at that time I wasn’t aware of the Contributor Guidelines. From now on, I’ll make sure to follow the Contributor Guidelines for all my contributions. Thank you for pointing this out.

PR #1337, opened 2026-03-02
Fix potential infinite loop in get_dependent_results_from_database

mrhapile:
## Proposed change

Fix potential infinite wait in get_dependent_results_from_database.

The polling loop inside this function could block indefinitely if
dependent module results were never written to the database.
In such cases, the scan engine could hang and never return control.

This change adds a 30-second timeout safeguard to prevent infinite blocking.

If no dependent event is found within the timeout window:
- A warning is logged
- The function exits safely
- The engine continues execution without deadlock

No architectural changes were introduced.
No function signatures were modified.
Existing behavior remains unchanged when events are present.

## Type of change

## Checklist

### DIFF
--- a/docs/Modules.md
+++ b/docs/Modules.md
@@ -34,6 +34,7 @@ OWASP Nettacker Modules can be of type **Scan** (scan for something), **Vuln** (
 - '**moveit_version_scan**' - Scan the target and identify the Progress MOVEit version
 - '**pma_scan**' - Scan the target for PHP MyAdmin presence
 - '**port_scan**' - Scan the target for open ports identifying the popular services using signatures (.e.g SSH on port 2222)
+- '**smartermail_version_scan**' - Scan the target and identify the SmarterMail version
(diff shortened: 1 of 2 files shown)


Aarush289:
Hi @mrhapile , this is not a valid issue , steps are first sorted in appropriate manner to take care that independent steps are executed first and then dependent, you may go through the sort_loop functions in modules.py to get more clarity .

mrhapile:
Hi @Aarush289 

My concern was around cases where a dependent event might never be written to the database (e.g., unexpected module failure), which could cause the polling loop to wait indefinitely.
If the current design guarantees that this situation cannot occur, I’m happy to close the PR. Could you please confirm whether such a safeguard already exists?

Aarush289:
If a module crashes then it will terminate the program flow and throw an exception , in that case it won't execute the dependent steps . Also it's better if you can create a situation when the mentioned error happens rather than saying possible error .

mrhapile:
That makes sense , I was considering defensive handling, but I understand that under the current execution model a dependent step would not be reached if a module crashes.

PR #1370, opened 2026-03-08
Render HTML scan reports inline in browser instead of downloading

### LINKED ISSUE #1316: Scan reports in HTML are downloaded in WebUI
# Summary

Currently, when using the WebUI, clicking the link for a completed scan result triggers an automatic download of the HTML file. While having the file locally is useful for archiving, it would improve the user experience if the report rendered directly in the browser.

# Steps to Reproduce

In WebUI, click any scan result e.g. `https://127.0.0.1:8080/results/get?id=2`, the result in HTML is downloaded

Image

I expected scan results in HTML are viewable in the browser window of WebUI 

Image

# Suggested Fix

The API in `nettacker/api/engine.py` can change `text/plain` to `text/html` and `attachment` to `inline` at least for HTML files.

https://github.com/OWASP/Nettacker/blob/ed9c8cad306fdf96b26b54175c30ab218bb84851/nettacker/api/engine.py#L373-L376

https://github.com/OWASP/Nettacker/blob/ed9c8cad306fdf96b26b54175c30ab218bb84851/nettacker/api/engine.py#L393-L394

I am curious if the current "force download" behavior was a deliberate design choice to mitigate security risks, such as preventing potential Cross-Site Scripting (XSS). We might implement a secure rendering method like a sandboxed iframe or strict Content-Security-Policy headers to balance safety with a smoother UI.

I would love to contribute a fix for this if the maintainers agree this is a worthwhile improvement. Please let me know if you would like me to submit a PR.

Pbhacks:
## Proposed change

Currently, clicking a scan result link in the WebUI (e.g. `https://127.0.0.1:8080/results/get?id=2`) triggers an automatic download of the HTML file instead of rendering it in the browser. This degrades user experience.

This PR fixes the `/results/get` endpoint so HTML reports render inline in the browser, with hardened security headers.

### Changes

`nettacker/api/engine.py` -- `/results/get` endpoint:

- Changed `Content-Disposition` from `attachment` to `inline` for `.html`/`.htm` files
- Added `sandbox allow-scripts` CSP directive (permits D3.js/jQuery visualizations, omits `allow-same-origin` to isolate origin)
- Added `base-uri 'none'` and `frame-ancestors 'none'` CSP directives
- Used `secure_filename(os.path.basename())` for safe filename handling with empty-string fallback
- Quoted filename in `Content-Disposition` per RFC 6266
- Handle `None` return from `get_scan_result()` with 404 response
- Handle `dict` return from `get_scan_result()` with 500 response
- Added `log.error()` on database exceptions for debuggability
- Non-HTML files continue to download as before (`attachment`)

`nettacker/database/db.py` -- `get_scan_result()`:

- Normalized APSW no-row case to return `None` instead of `structure(dict)`, consistent with SQLAlchemy branch (both yield 404 now)

`nettacker/api/core.py` -- `mime_types()`:

- Fixed incorrect `.htm` mime type (was `'.htm'` instead of `'text/html'`)

`tests/api/test_engine.py` (new):

- 7 endpoint-level regression tests for `/results/get`: inline HTML+CSP, .htm variant, non-HTML attachment, 404 not-found, 500 dict-error, 400 missing-id, empty safe_filename fallback

`tests/api/test_core.py`:

- Added `.htm` assertion to `test_mime_types` to prevent regressions

Fixes #1316

## Type of change

## Checklist

- [x] I've followed the [contributing guidelines](https://nettacker.readthedocs.io/en/latest/Developers/)
- [x] I've run `make pre-commit` and confirm it didn't generate any warnings/changes
- [x] I've run `make test`, I confirm all tests passed locally

### DIFF
--- a/.github/workflows/ci_cd.yml
+++ b/.github/workflows/ci_cd.yml
@@ -149,7 +149,7 @@ jobs:
           python-version: '3.11'
 
       - name: Get package artifacts
-        uses: actions/download-artifact@70fc10c6e5e1ce46ad2ea6f2b72d43f7d47b13c3
+        uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c
         with:
           name: dist
           path: dist
@@ -338,7 +338,7 @@ jobs:
     runs-on: ubuntu-24.04
     steps:
       - name: Get package artifacts
-        uses: actions/download-artifact@70fc10c6e5e1ce46ad2ea6f2b72d43f7d47b13c3
+        uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c
         with:
           name: dist
           path: dist
@@ -364,7 +364,7 @@ jobs:
     runs-on: ubuntu-24.04
     steps:
       - name: Get package artifacts
-        uses: actions/download-artifact@70fc10c6e5e1ce46ad2ea6f2b72d43f7d47b13c3
+        uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c
         with:
           name: dist
           path: dist


nettacker/api/engine.py
```diff
@@ -388,10 +388,28 @@ def get_result_content():
...
+    headers = {}
+
+    if ext in (".html", ".htm"):
+        headers["Content-Disposition"] = "inline;filename=" + safe_filename
+        headers["Content-Security-Policy"] = (
+            "default-src 'none'; "
+            "style-src 'self' 'unsafe-inline'; "
+            "script-src 'self' 'unsafe-inline'; "
+            "img-src 'self' data:; "
+            "font-src 'self' data:"
```
jess-tech-lab: I suggest adding a sandbox attribute.
```
"sandbox allow-scripts; "
```

By using `sandbox allow-scripts`, we permit the D3.js and jQuery visualizations to render correctly. By omitting `allow-same-origin`, we force the report into a unique origin, which prevents any potentially malicious scan results from accessing site-wide cookies or local storage. This helps to harden the report without breaking the interactive features.

jess-tech-lab:
```suggestion
            "font-src 'self' data:; "
```

Pbhacks: Sure

Pbhacks: Done

securestep9:
@coderabbitai remember to check for: 1. signed commits. 2. conflicts with other open pull requests in this repo
@coderabbitai resume

PR #1376, opened 2026-03-08
fix(cors): resolve regex false negative and add value-based regex coverage

### LINKED ISSUE #1374: Silent failure in http_cors.yaml due to unvalidated regex logic
# Summary

There is a "silent failure" where the `http_cors.yaml` module fails to detect vulnerabilities due to an incorrect regex pattern with a trailing space. Currently, the test suite (`tests/test_yaml_regexes.py`) only checks if a regex is syntactically valid, not if it actually matches the expected data.

As a result, `http_cors.yaml` passes all tests despite being logically broken for real-world scanning.

# Description
In `nettacker/modules/vuln/http_cors.yaml`, the regex `(http|https):\\/\\/evil.com ` contains a trailing space.
https://github.com/OWASP/Nettacker/blob/2310a8398f788bf7c3b767a153da151ab2a49445/nettacker/modules/vuln/http_cors.yaml#L336-L337

- Expected behavior: Match `https://evil.com`
- Actual behavior: Fails to match because servers do not append a space to the `Access-Control-Allow-Origin` header.

Testing gap: The current `tests/test_yaml_regexes.py` sees this as a valid regex string and passes it, even though it will never trigger a "Detected" status in a real scan.

Shirshaw64p:
## Summary
Fixes silent failure in `http_cors.yaml` where a trailing space in regex prevented matching reflected origins.

## Changes
- Fixed trailing-space regex in `nettacker/modules/vuln/http_cors.yaml`
- Expanded `tests/test_yaml_regexes.py` to validate all HTTP `response.conditions` regexes (including headers)
- Added regression test to assert `https://evil.com` matches the CORS reflected-origin regex
- Updated YAML test loader to `encoding="utf-8"` for reliable cross-platform parsing

Closes #1374

### DIFF
--- a/.github/workflows/ci_cd.yml
+++ b/.github/workflows/ci_cd.yml
@@ -54,12 +54,12 @@ jobs:
       - name: Check out repository
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd
       - name: Initialize CodeQL
-        uses: github/codeql-action/init@c793b717bc78562f491db7b0e93a3a178b099162
+        uses: github/codeql-action/init@0d579ffd059c29b07949a3cce3983f0780820c98
         with:
           languages: ${{ matrix.language }}
 
       - name: Perform CodeQL Analysis
-        uses: github/codeql-action/analyze@c793b717bc78562f491db7b0e93a3a178b099162
+        uses: github/codeql-action/analyze@0d579ffd059c29b07949a3cce3983f0780820c98
         with:
           category: /language:${{ matrix.language }}
 
@@ -280,7 +280,7 @@ jobs:
           platforms: linux/amd64,linux/arm64
 
       - name: Build and push
-        uses: docker/build-push-action@10e90e3645eae34f1e60eeb005ba3a3d33f178e8
+        uses: docker/build-push-action@d08e5c354a6adb9ed34480a06d141179aa583294
         with:
           context: .
           file: Dockerfile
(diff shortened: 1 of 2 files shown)


Shirshaw64p:
Addressed CodeRabbit/Copilot feedback in latest commits.
All local targeted tests pass (tests/test_yaml_regexes.py).
Could a maintainer please approve workflows and review/sign off?

Also, I can re-push with verified signed commits if required.

tests/test_yaml_regexes.py
```diff
@@ -85,3 +101,21 @@ def test_yaml_regexes_valid(yaml_file):
 
     for regex in regexes:
         assert is_valid_regex(regex), f"Invalid regex in {yaml_file}: `{regex}`"
+
+
+def test_http_cors_reflected_origin_regex_matches_expected_origin():
```
pUrGe12, line 106: While more tests are generally good, I am not sure if this way of testing is extendable to all modules. Can you not extend the `DUMMY_TEST_STRING` we already have with your cases and match on that? Do we need a separate function here?

tests/test_yaml_regexes.py
```diff
@@ -28,12 +28,28 @@ def extract_http_regexes(payloads):
...
         for step in payload.get("steps", []):
-            conditions = step.get("response", {}).get("conditions", {})
-            if "content" in conditions and "regex" in conditions["content"]:
-                regexes.append(conditions["content"]["regex"])
+            response = step.get("response", {})
+            regexes.extend(extract_regex_values(response))
     return regexes
 
 
+def extract_regex_values(response):
```
pUrGe12, line 36: Won't the current `extract_socket_regexes` and `extract_http_regexes` do? Why do you need a separate function here?

pUrGe12:
Also, potential duplicate of #1375? Are you not happy with the fix in that PR?

Shirshaw64p:
I saw that now. Please proceed with that one. :-D

PR #1409, opened 2026-03-15
remove exec() from responsetime condition evaluator

### LINKED ISSUE #1408: Arbitrary code execution via exec() in response_conditions_matched
Summary
response_conditions_matched() in nettacker/core/lib/http.py evaluates the responsetime condition using exec() with an operator taken directly from module YAML.

This allows arbitrary Python execution if a crafted module provides a malicious condition string.

Location:
nettacker/core/lib/http.py
~line 95
Vulnerable pattern
exec(
    f"matched = response['responsetime'] "
    f"{conditions['responsetime'].split()[0]} "
    f"{conditions['responsetime'].split()[1]}"
)

Since the operator comes from YAML, a condition like:

conditions:
  responsetime: ">= 0 or __import__('os').system('id')"

would execute arbitrary code during scan.

Fix

Replace exec() with explicit operator dispatch:

ops = {
    '==': float.__eq__,
    '!=': float.__ne__,
    '>=': float.__ge__,
    '<=': float.__le__,
    '>':  float.__gt__,
    '<':  float.__lt__,
}

Gauravsharma2040:
This PR removes the use of exec() in
nettacker/core/lib/http.py → response_conditions_matched()
when evaluating the responsetime condition.

The previous implementation constructed a Python expression using
values taken directly from module YAML and executed it with exec(),
which allows arbitrary code execution if a crafted condition string is provided.

Example of previous pattern:

exec(
    f"matched = response['responsetime'] "
    f"{conditions['responsetime'].split()[0]} "
    f"{conditions['responsetime'].split()[1]}"
)

Since the operator comes from YAML, this is unsafe.

Changes

Replace exec() with explicit operator dispatch

Allow only valid comparison operators:

== != > < >= <=

Preserve existing behaviour for valid modules

No API / module format changes

New logic uses a safe mapping:

ops = {
    '==': float.__eq__,
    '!=': float.__ne__,
    '>=': float.__ge__,
    '<=': float.__le__,
    '>':  float.__gt__,
    '<':  float.__lt__,
}  
fixes #1408

### DIFF
--- a/nettacker/core/lib/http.py
+++ b/nettacker/core/lib/http.py
@@ -27,8 +27,8 @@ _RESPONSETIME_OPS = {
     "!=": operator.ne,
     ">=": operator.ge,
     "<=": operator.le,
-    ">":  operator.gt,
-    "<":  operator.lt,
+    ">": operator.gt,
+    "<": operator.lt,
 }
 
 


securestep9:
@coderabbitai remember to check all other open pull requests which may conflict with this one

Gauravsharma2040:
@securestep9  
CodeRabbit suggested keeping this PR open because of the additional input guards and type checks, but both approaches seem functionally valid.
Would you prefer that I close this PR in favor of the existing one , or keep this open ?
Happy to update or close this PR depending on what fits best with the current review.

PR #1414, opened 2026-03-16
Add CVE-2022-22947 Spring Cloud Gateway vulnerability detection module

Prajwal5755:
## Proposed change

This PR adds a vulnerability detection module for **CVE-2022-22947 (Spring Cloud Gateway)**.

The module confirms the vulnerability by reproducing the exploitation flow:

1. Injects a new route using the actuator endpoint  
   `/actuator/gateway/routes/nettacker-test`

2. Refreshes the gateway configuration using  
   `/actuator/gateway/refresh`

3. Verifies the injected route via  
   `/actuator/gateway/routes/nettacker-test`

If the injected route is successfully created and returned, the module confirms the vulnerability.

Tested locally using the **Vulhub CVE-2022-22947 lab environment**.

## Type of change

- [ ] Dependency upgrade

## Checklist

- [x] I have **digitally signed** all my commits in this PR
- [ ] I've run `make pre-commit` and confirm it didn't generate any warnings/changes
- [ ] I've run `make test`, I confirm all tests passed locally
- [ ] I've linked this PR with an open issue
- [x] I have attached screenshots demonstrating my code works as intended

Screenshot from 2026-03-17 06-31-32

### DIFF
--- a/nettacker/modules/vuln/spring_cloud_gateway_cve_2022_22947.yaml
+++ b/nettacker/modules/vuln/spring_cloud_gateway_cve_2022_22947.yaml
@@ -1,0 +1,130 @@
+info:
+  name: spring_cloud_gateway_cve_2022_22947_vuln
+  author: Prajwal G N (@Prajwal5755)
+  severity: 9.8
+  description: Detects Spring Cloud Gateway actuator route injection vulnerability (CVE-2022-22947).
+  reference:
+    - https://nvd.nist.gov/vuln/detail/CVE-2022-22947
+    - https://spring.io/security/cve-2022-22947
+    - https://tanzu.vmware.com/security/cve-2022-22947
+  profiles:
+    - vuln
+    - http
+    - critical_severity
+    - cve
+    - spring
+
+payloads:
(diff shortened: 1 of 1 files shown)


Aarush289:
Hi @Prajwal5755  , in the module , irrespective of the first and second step , if the third one is successful it will give success , this is because of how save to temp events work , it will always store the response of that step in temporary database even if the regex matching conditions fail so if you are using it in some other step just to check whether the previous one was successful or not then it won't work as intentional , it will always pass that check even if the previous step was unsuccessful. Please correct me if you think I am wrong at any step

Prajwal5755:
Thanks for pointing this out  @Aarush289 .
You are right that `save_to_temp_events_only` stores events even if conditions fail, so `dependent_on_temp_event` alone does not guarantee correctness.
To address this, I ensured that the final verification step strictly checks for the injected marker (`nettacker-test`) in the response content. This confirms that the route injection actually succeeded rather than relying only on step execution.
Please let me know if you would  suggest any further improvements.

PR #1420, opened 2026-03-18
Fix runtime bugs , redundancies and duplicates in web UI (main.js)

### LINKED ISSUE #1419: Web UI main.js contains multiple runtime bugs and some redundancies in crawler view and pagination
While working on the web UI I noticed several small runtime bugs and code quality issues in
nettacker/web/static/js/main.js that affect the crawler view, pagination, and compare view.

These issues do not affect the scan engine or API, but they cause incorrect UI behaviour.

Observed problems

1. Duplicate #compare_btn click handler  
   Two handlers are bound to the same button.  
   The first runs without session validation, so the compare panel can appear briefly
   before the AJAX check hides it.

2. Pagination uses string length instead of result count  
   In get_crawler_list(), total pages are calculated using res.length while res is
   still a JSON string. This produces incorrect page counts for larger datasets.

3. totalPages is scoped inside callback  
   totalPages is declared inside the AJAX .done() block but is read by button handlers,
   causing last-page navigation to fail.

4. setTimeout used with string arguments  
   Multiple places use string-based setTimeout which relies on implicit eval and
   prevents animations from working correctly.

5. Pagination button handlers capture loop variable incorrectly  
   All page buttons navigate to the last page due to closure over the loop variable.

6. Duplicate getUrlParameter definition  
   Function is declared twice, second definition overwrites the first.

7. Redundant toggle calls in pagination controls  
   Buttons are toggled multiple times in different places, making logic harder to follow.

Proposed fix
I would like to submit a single PR that fixes these issues in main.js only.

Scope of the proposed PR:
- File: nettacker/web/static/js/main.js
- No backend changes
- No API changes
- No scan logic changes
- No new features, only bug fixes and cleanup

Please let me know if it is preferred if these fixes are split into multiple PRs.

Gauravsharma2040:
This pr addresses several runtime bugs and code hygiene issues in

nettacker/web/static/js/main.js

These affect crawler view, results view, compare view, and general UI behaviour.
All problems are addressed within the scope of this pr are frontend-only and do not affect the scan engine or API.

Observed issues

- Duplicate #compare_btn click handler without session check
- Duplicate getUrlParameter declaration
- setTimeout called with string arguments (implicit eval)
- Fade animations removed before timeout executes
- Duplicate toggle() calls in updatePaginationControls
- Pagination button closures capturing shared loop variable
- filter_large_content scanning from wrong offset and returning early
- Implicit global variables (filter_index, sm, parts, etc.)
- Missing </a> tag in show_scans card HTML
- Stray quote in button class attributes
- console.log left in production code
- Dead / commented-out code blocks in show_scans and show_crawler
- Undefined passed to filter_large_content when "conditions:" not present
- DB-sourced values inserted into HTML without escaping (XSS risk)

Proposed fix

Single PR limited to:

- nettacker/web/static/js/main.js
- bug fixes
- removal of dead / commented code
- variable scope fixes
- safe string handling
- no UI redesign
- no backend changes

Out of scope (intentionally not modified)

- Pagination total count logic
- /logs/search API response format
- crw_last_btn / totalPages backend mismatch

These are being handled separately by a backend change that adds total_count
to the API response, so this PR will not modify pagination logic.

Goal
fixes #1419

### DIFF
--- a/nettacker/web/static/js/main.js
+++ b/nettacker/web/static/js/main.js
@@ -187,7 +187,7 @@ $(document).ready(function () {
         } else {
           document.getElementById("report_error_msg").innerHTML = response.message;
           $("#failed_report").removeClass("hidden");
-          setTimeout(function () { $("#failed_report").addClass("hidden"); }, 5000); // FIX #3
+          setTimeout(function () { $("#failed_report").addClass("hidden"); }, 5000); 
         }
       })
       .fail(function (jqXHR, textStatus, errorThrown) {
@@ -403,7 +403,7 @@ $(document).ready(function () {
         document.getElementById("error_msg").innerHTML = jqXHR.responseText;
         if (errorThrown === "BAD REQUEST" || errorThrown === "UNAUTHORIZED") {
           $("#failed_request").removeClass("hidden");
-          setTimeout(function () { $("#failed_request").addClass("hidden"); }, 5000); // FIX #3
+          setTimeout(function () { $("#failed_request").addClass("hidden"); }, 5000); 
         }
       });
   });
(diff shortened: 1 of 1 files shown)


Gauravsharma2040:
Thanks for the review @coderabbitai. To clarify the scope of this pr none of the three issues flagged originate from my changes. I've verified each against the pre-existing file and will update the pr for fixing these as well wherever a trivial fix is possible.

Gauravsharma2040:
@coderabbitai also check for conflicts with existing PRs please

Gauravsharma2040:
@coderabbitai Documenting:  I will remove fixing crawler results pagination from the scope of this pr since full rectification might require changes to other files out of current scope This pr only fixes/cleans bugs within main.js itself will push a commit with fixes for newly addressed issues and crawler pagination UI untouched

PR #1426, opened 2026-03-19
Hide pagination controls when no results exist in Crawler page

mdtaha-dev:
## Proposed change

The Crawler page in the Nettacker web UI incorrectly displays pagination 
buttons (Previous, Next, and numbered page buttons) even when there are 
no scan results. This fix hides pagination controls when the result set 
is empty and only shows them when actual data exists.

## Problems

1. Incorrect totalPages calculation
   * Response length was being calculated on the raw JSON string instead of the parsed array
   * This caused totalPages to always be a large number regardless of actual results
   * Pagination buttons were always rendered even with zero results

2. Poor empty state UX
   * Users saw 5 page buttons and a Next button on an empty Crawler page
   * This was confusing and suggested data existed when it did not
   * No visual feedback that the page was actually empty

## Proposed solution

**Frontend (main.js)**

Parse the JSON response before calculating total pages:
* Use `JSON.parse(res)` before `Math.ceil(parsedRes.length / 10)`
* Check `parsedRes.length === 0` before rendering pagination
* Hide all pagination controls when result set is empty
* Only call `updatePaginationControls()` when results exist

This keeps the fix minimal and non-breaking. Existing pagination logic 
for non-empty results is completely unchanged.

## Scope

This change only affects the Crawler page pagination logic in main.js. 
No changes to backend, scan engine, or any other UI component.

### DIFF
--- a/.github/workflows/ci_cd.yml
+++ b/.github/workflows/ci_cd.yml
@@ -54,12 +54,12 @@ jobs:
       - name: Check out repository
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd
       - name: Initialize CodeQL
-        uses: github/codeql-action/init@b1bff81932f5cdfc8695c7752dcee935dcd061c8
+        uses: github/codeql-action/init@38697555549f1db7851b81482ff19f1fa5c4fedc
         with:
           languages: ${{ matrix.language }}
 
       - name: Perform CodeQL Analysis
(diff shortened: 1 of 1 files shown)


Gauravsharma2040:
@coderabbitai could confirm if this pr conflicts with others prs like #1294  , #1420 .

securestep9:
@coderabbitai full review

PR #1430, opened 2026-03-20
Fix: Render HTML scan reports inline instead of downloading

### LINKED ISSUE #1316: Scan reports in HTML are downloaded in WebUI

Devanshshar01:
## Summary

Fixes #1316

Currently, clicking a scan result link in the WebUI (e.g. `/results/get?id=2`) triggers an automatic file download of the HTML report. This PR changes that so HTML reports open and render directly in the browser.

## Root Cause

[get_result_content()](cci:1://file:///c:/Users/User/Desktop/Nettacker/nettacker/api/engine.py:372:0-411:5) in [nettacker/api/engine.py](cci:7://file:///c:/Users/User/Desktop/Nettacker/nettacker/api/engine.py:0:0-0:0) was unconditionally setting `Content-Disposition: attachment` for all file types, forcing the browser to download instead of render.

## Changes

**Modified:** [nettacker/api/engine.py](cci:7://file:///c:/Users/User/Desktop/Nettacker/nettacker/api/engine.py:0:0-0:0)

- For `.html` / `.htm` files:
  - `Content-Disposition` changed from [attachment](cci:1://file:///c:/Users/User/Desktop/Nettacker/tests/api/test_engine.py:155:0-167:29) to [inline](cci:1://file:///c:/Users/User/Desktop/Nettacker/tests/api/test_engine.py:121:0-133:33)
  - `mimetype` set explicitly to `text/html`
  - Added `Content-Security-Policy: default-src 'self'; script-src 'none'; style-src 'self' 'unsafe-inline'` to mitigate XSS risk
- For all other file types (`.json`, [.txt](cci:7://file:///c:/Users/User/Desktop/Nettacker/nettacker/logo.txt:0:0-0:0), `.csv`): **behaviour unchanged** — still downloaded as attachments

**Added:** [tests/api/test_engine.py](cci:7://file:///c:/Users/User/Desktop/Nettacker/tests/api/test_engine.py:0:0-0:0)

New unit tests for the `/results/get` endpoint:
- [test_html_report_rendered_inline](cci:1://file:///c:/Users/User/Desktop/Nettacker/tests/api/test_engine.py:103:0-118:83) — HTML uses [inline](cci:1://file:///c:/Users/User/Desktop/Nettacker/tests/api/test_engine.py:121:0-133:33), `text/html`, and CSP blocks scripts
- [test_htm_report_rendered_inline](cci:1://file:///c:/Users/User/Desktop/Nettacker/tests/api/test_engine.py:121:0-133:33) — `.htm` behaves the same as `.html`
- [test_json_report_downloaded_as_attachment](cci:1://file:///c:/Users/User/Desktop/Nettacker/tests/api/test_engine.py:136:0-152:41) — JSON still downloads
- [test_txt_report_downloaded_as_attachment](cci:1://file:///c:/Users/User/Desktop/Nettacker/tests/api/test_engine.py:155:0-167:29) — TXT still downloads

## Security Consideration

Rendering HTML inline introduces potential XSS risk if report content is adversarial. This is mitigated by setting `script-src 'none'` in the Content-Security-Policy, which blocks all script execution. Nettacker HTML reports are static styled documents that do not require JavaScript.

## Testing

- All 4 new unit tests pass
- `ruff check` linting passes cleanly
- Non-HTML types verified to still download via tests

### DIFF
--- a/nettacker/api/engine.py
+++ b/nettacker/api/engine.py
@@ -393,17 +393,26 @@ def get_result_content():
     safe_filename = secure_filename(os.path.basename(str(filename)))
 
     # Render HTML reports inline in the browser instead of forcing a download.
-    # Non-HTML files continue to be served as attachment downloads.
+    # This allows users to view scan results directly while maintaining security.
+    # Non-HTML files (.json, .csv, .txt) continue to be served as attachment downloads.
     if file_extension in (".html", ".htm"):
         response = Response(
             file_content,
             mimetype="text/html",
-            headers={"Content-Disposition": "inline;filename=" + safe_filename},
+            headers={"Content-Disposition": f"inline; filename={safe_filename}"},
         )
-        # Restrict inline HTML content to mitigate XSS risks (no scripts, no external resources)
-        response.headers[
-            "Content-Security-Policy"
-        ] = "default-src 'self'; script-src 'none'; style-src 'self' 'unsafe-inline'"
+        # Security: Strict Content-Security-Policy to mitigate XSS risks.
+        # We disable all script execution (script-src 'none') and restrict other resources
+        # to ensure that user-generated HTML reports cannot execute malicious code.
+        response.headers["Content-Security-Policy"] = (
+            "default-src 'none'; "
+            "script-src 'none'; "
+            "style-src 'self' 'unsafe-inline'; "
+            "img-src 'self' data:; "
+            "font-src 'self'; "
+            "connect-src 'self'; "
+            "frame-ancestors 'none';"
+        )
         return response
 
     return Response(
--- a/tests/api/test_engine.py
+++ b/tests/api/test_engine.py
@@ -100,7 +100,7 @@ def _stub_nettacker_modules():
     # Expose it on the fixture's module so tests can reference it.
     _stub_nettacker_modules.app = _app  # type: ignore[attr-defined]
 
-    yield
+    yield _stub_nettacker_modules
 
     # Restore sys.modules to its original state.
     for name, original in originals.items():
@@ -145,7 +145,9 @@ def test_html_report_rendered_inline(mock_get_scan_result, mock_api_key, client)
     assert "attachment" not in cd, "attachment must NOT appear for HTML reports"
     assert "report.html" in cd
     csp = response.headers.get("Content-Security-Policy", "")
+    assert "default-src 'none'" in csp
     assert "script-src 'none'" in csp, f"Expected CSP to block scripts, got: {csp}"
+    assert "frame-ancestors 'none'" in csp
 
 
 @patch("nettacker.api.engine.api_key_is_valid")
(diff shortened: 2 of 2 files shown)


Devanshshar01:
Thanks for the review! I've addressed all the suggested changes and updated the PR accordingly. Please let me know if anything else needs improvement.

securestep9:
@coderabbitai remember to run your checks and learnings on each PR

Devanshshar01:
Hi! I’ve updated the PR:

- Signed commits using GPG
- Fixed pre-commit/CI issues
- Addressed all review comments

Please let me know if any further improvements are needed. Thanks!

Devanshshar01:
Hi! Just to clarify, this implementation ensures:
- Secure filename handling
- Strict CSP to mitigate XSS risks
- Support for both .html and .htm

Happy to simplify further if preferred.

Devanshshar01:
Hi @securestep9, I’ve addressed the previous feedback and updated the PR accordingly. 

I’d really appreciate it if you could take a look when you have time. Thanks!

pUrGe12:
I have a question: why? What's the problem with downloading the html content? you can always open that in a browser later. We don't have to worry about the security risks that you're mentioning then

Devanshshar01:
@pUrGe12 
Thanks for bringing this up.

The main idea behind rendering HTML reports inline instead of forcing a download is to improve both usability and safety.

From a usability standpoint, it lets users instantly view scan results in the browser without the extra step of downloading and opening the file manually. This makes the workflow smoother, especially when running multiple scans.

On the security side, even downloaded HTML files can still execute scripts when opened locally in a browser. By serving them inline with a strict Content Security Policy (e.g., blocking scripts), we can ensure the content is rendered in a safer way.

Also, this approach is consistent with how many tools display reports/logs directly in the browser rather than forcing downloads.

That said, I’m open to adjusting the behavior if download-by-default is preferred — just went with inline as a more user-friendly default.

Happy to refine this further based on your thoughts.

PR #1435, opened 2026-03-21
Optimization: Pre-emptive Module Filtration via Service Discovery

D0NTASKME:
Implements service-aware task allocation to reduce thread overhead and eliminate SQLite database lock contention.

## Proposed change

Currently, the engine allocates threads for every selected module regardless of whether the required service (HTTP, FTP, SSH, etc.) exists on the target. This results in $O(N \times M)$ thread overhead and massive SQLite lock contention.

This PR introduces Pre-emptive Module Filtration. By checking discovered services from the port_scan phase against module requirements before thread spawning, we eliminate redundant resource allocation.

## Type of change

- [ ] Dependency upgrade
- [x] Other improvement (best practice, cleanup, optimization, etc)

## Checklist

- [x] I have **digitally signed** all my commits in this PR
- [x] I've run `make pre-commit` and confirm it didn't generate any warnings/changes
- [x] I've run `make test`, I confirm all tests passed locally
- [x] I have attached screenshots demonstrating my code works as intended
 
Any feedback to further optimise my solution would be greatly appreciated!
Some tests showing comparisons of time taken on scan running with and without code  I have added, to stress test:

**Before code:**
image
image

**After code additions:**
image
image

### DIFF
--- a/.github/workflows/ci_cd.yml
+++ b/.github/workflows/ci_cd.yml
@@ -54,12 +54,12 @@ jobs:
       - name: Check out repository
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd
       - name: Initialize CodeQL
-        uses: github/codeql-action/init@b1bff81932f5cdfc8695c7752dcee935dcd061c8
+        uses: github/codeql-action/init@38697555549f1db7851b81482ff19f1fa5c4fedc
         with:
           languages: ${{ matrix.language }}
 
       - name: Perform CodeQL Analysis
-        uses: github/codeql-action/analyze@b1bff81932f5cdfc8695c7752dcee935dcd061c8
+        uses: github/codeql-action/analyze@38697555549f1db7851b81482ff19f1fa5c4fedc
         with:
           category: /language:${{ matrix.language }}
 
--- a/docs/Usage.md
+++ b/docs/Usage.md
@@ -1,21 +1,19 @@
 # Help Menu
 
 - [Target inputs Option](#target-inputs-option)
-  * [Command Examples](#command-examples)
+  - [Command Examples](#command-examples)
 - [API and WebUI](#api-and-webui)
-  * [API Options](#api-options)
-  * [API Examples](#api-examples)
+  - [API Options](#api-options)
+  - [API Examples](#api-examples)
 - [Database](#database)
-  * [SQLite configuration](#sqlite-configuration)
-  * [MySQL configuration](#mysql-configuration)
+  - [SQLite configuration](#sqlite-configuration)
+  - [MySQL configuration](#mysql-configuration)
 - [Maltego Transforms](#maltego-transforms)
 
 By using the `--help`/`-h` switch you can read the help menu in the CLI:
- `python3 nettacker.py --help`
+`python3 nettacker.py --help`
 
-
-
-* Note: This example may not reflect the latest version.
+- Note: This example may not reflect the latest version.
 
 ```
    ______          __      _____ _____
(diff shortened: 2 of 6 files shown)


nettacker/core/app.py
```diff
@@ -194,6 +195,77 @@ def filter_target_by_event(self, targets, scan_id, module_name):
...
+            return True
+        if cache is not None and target in cache["targets"]:
+            discovered_libraries = cache["targets"][target]
+        else:
+            found_events = find_events(target, "port_scan", scan_id)
+            discovered_libraries = set()
+            for event in found_events:
+                event_data = json.loads(event)
+                libraries = event_data.get("response", {}).get("conditions_results", {})
+                discovered_libraries.update(libraries.keys())
```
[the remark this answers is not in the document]
D0NTASKME: Fixed by referencing services and libraries both

D0NTASKME:
@arkid15r @securestep9 - Hi! CI/CD workflow needs maintainer approval to run tests. 
Could you approve the workflow? Thanks!

nettacker/core/app.py
```diff
@@ -194,6 +195,79 @@ def filter_target_by_event(self, targets, scan_id, module_name):
...
+        else:
+            inputs = self.arguments.__dict__.copy()
+            inputs["target"] = target
+            # Get the module's requirements (e.g., 'http')
+            module_content = TemplateLoader(module_name, inputs).load()
+            required_services = set()
+            for payload in module_content.get("payloads", []):
+                service = payload.get("service") or payload.get("library")
+                if service:
+                    required_services.add(service)
```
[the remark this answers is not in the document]
D0NTASKME: Fixed by adding try-except block in latest commit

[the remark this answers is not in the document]
D0NTASKME: Fixed by adding try-except block in latest commit

nettacker/core/app.py
```diff
@@ -292,9 +366,12 @@ def scan_target_group(self, targets, scan_id, process_number):
         log.verbose_event_info(_("single_process_started").format(process_number))
         total_number_of_modules = len(targets) * len(self.arguments.selected_modules)
         total_number_of_modules_counter = 1
-
+        scan_cache = {"modules": {}, "targets": {}}
         for target in targets:
             for module_name in self.arguments.selected_modules:
+                if not self.is_module_applicable(target, module_name, scan_id, cache=scan_cache):
+                    log.info(f"Skipping {module_name} for {target}: Service not found.")
+                    continue
```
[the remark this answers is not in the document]
D0NTASKME: Generated key and signed recent commits.

D0NTASKME:
@arkid15r @securestep9 Following up—I've now fully resolved the CodeRabbit feedback:
​-  Added try/except blocks to is_module_applicable to ensure the engine "fails open" on corrupted templates/events.
​ - Re-signed all commits with GPG (Verified badge).
​- Fixed the service vs. library mapping mismatch caught in the analysis.
​Ready for the CI/CD run whenever you have a moment to approve the workflow. Thanks!

nettacker/core/app.py
```diff
@@ -194,6 +195,85 @@ def filter_target_by_event(self, targets, scan_id, module_name):
...
+            required_services = set()
+            for payload in module_content.get("payloads", []):
+                service = payload.get("service") or payload.get("library")
+                if service:
+                    required_services.add(service)
+            if cache is not None:
+                cache["modules"][module_name] = required_services
+        if not required_services:
+            return True
+        port_scan_vocabulary = {
```
pUrGe12: This is not good here. Any addition to the service discovery regexes will require people to come here and change this set inside main.

nettacker/core/app.py
```diff
@@ -194,6 +195,85 @@ def filter_target_by_event(self, targets, scan_id, module_name):
...
+            cache: (Optional) A dictionary for caching module requirements 
+                   and target services to optimize performance.
+
+        Returns:
+            bool: True if the module is applicable or unknown, 
+                  False if the service mismatch is confirmed.
+        """
+        if self.arguments.skip_service_discovery:
+            return True
+        if module_name in ["port_scan", "subdomain_scan", "icmp_scan"]:
```
pUrGe12: Note that such a thing already happens in the same file. We split the checks based on these three module names, its redundant to do it again. Can you hook that in instead.

nettacker/core/app.py
```diff
@@ -26,6 +26,7 @@
 from nettacker.core.messages import messages as _
 from nettacker.core.module import Module
 from nettacker.core.socks_proxy import set_socks_proxy
+from nettacker.core.template import TemplateLoader
```
pUrGe12: Ideally all template loading and module parsing should go into module.py

pUrGe12:
^ my two cents on your PR

The database locking issue that you're seeing is probably due to the number of threads you were running (was is 125?). The default database handler (that is, for sqlite) has been changed to APSW which is a raw implementation over sqlite3 for lower overhead on database tasks.

If your system cannot handle 125 threads:

1. Try lower the number
2. Switch to the ORM (you can do this using the `config.py`: set `use_apsw_for_sqlite` to `False`). The ORM will ALSO lock your database if you're using SQLite because sqlite just doesn't support concurrent additions, it'll just manage resources better.

nettacker/core/app.py
D0NTASKME: Refactored code by extracting required services in module.py, as part of module class.

nettacker/core/app.py
D0NTASKME: I've removed the redundancy inside is_module_applicable and hooked these bypass checks directly into the caller loop in scan_target_group instead

nettacker/core/app.py
D0NTASKME: I've completely removed the dictionary and refactored the code, it now extracts the required profiles directly from the module's metadata (via a new get_required_services method in module.py).

D0NTASKME:
Thanks for the insight on the thread limits! I agree that high thread counts stress the SQLite I/O. However, the main goal of this PR is to stop ineligible modules before they even attempt a connection or a database write. By filtering them out pre-emptively, we eliminate thousands of redundant 'failed' transactions from being accessed by DB handler in the first place.

PR #1527, opened 2026-04-23
security: fix yaml.load deprecation and SQL injection in database creation

mitre88:
## Summary
- Replace deprecated `yaml.load()` with `yaml.safe_load()` to prevent arbitrary code execution through malicious YAML files
- Add database name validation to prevent SQL injection in CREATE DATABASE statements for both MySQL and PostgreSQL

## Changes
- `nettacker/core/messages.py`: Use `yaml.safe_load()` instead of `yaml.load()`
- `nettacker/database/mysql.py`: Add alphanumeric validation for database name and use backtick-quoted identifier
- `nettacker/database/postgresql.py`: Add alphanumeric validation for database name and use quoted identifier

## Security Impact
- `yaml.load()` with FullLoader can execute arbitrary Python code embedded in YAML files
- SQL interpolation in CREATE DATABASE statements could allow SQL injection if database name is user-controlled

### DIFF
--- a/nettacker/core/messages.py
+++ b/nettacker/core/messages.py
@@ -20,7 +20,8 @@ def application_language():
 
 
 def load_yaml(filename):
-    return yaml.safe_load(open(filename, "r"))
+    with open(filename, "r") as f:
+        return yaml.safe_load(f)
 
 
 def get_languages():
--- a/nettacker/core/messages.py
+++ b/nettacker/core/messages.py
@@ -20,7 +20,7 @@ def application_language():
 
 
 def load_yaml(filename):
-    return yaml.load(StringIO(open(filename, "r").read()), Loader=yaml.FullLoader)
+    return yaml.safe_load(open(filename, "r"))
 
 
 def get_languages():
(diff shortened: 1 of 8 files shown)


nettacker/core/messages.py
```diff
@@ -20,7 +20,8 @@ def application_language():
 
 
 def load_yaml(filename):
-    return yaml.load(StringIO(open(filename, "r").read()), Loader=yaml.FullLoader)
```
pUrGe12, line 23: This really isn't that big of a problem because we're reviewing each YAML file which goes in the codebase. That doesn't mean we can't use a safer implementation though.

pUrGe12 (changes requested):
The context manager for opening files is a good approach and you should probably keep that in a separate PR for better bookkeeping. For the security changes, as far as I am aware, the risks look real but most of these values are through the config, which I am assuming is set by an adminstrator when running the server or the user themselves while running locally. Obviously that doesn't mean we shouldn't secure the code while we have the chance. @securestep9 what do you think?

As for the database name, we cannot break backwards compatibility. The AI suggestion seems pretty important there.

mitre88:
Closing this to keep my open PR queue focused on contributions that are currently reviewable and likely to proceed. This one has a blocking signal (failing-checks) and would need a deeper rework before it is worth maintainer time. Thanks for the review/context.

PR #1590, opened 2026-06-21
fix(messages): use pathlib stem in get_languages for cross-platform language codes

### LINKED ISSUE #1582: [Bug] Windows compatibility issue in locale directory parsing (get_languages)
### Describe the bug

On Windows systems, the language discovery helper `get_languages()` fails to extract the short language codes (such as `en`, `fa`, `fr`) from the locale directory paths. Instead, it extracts the full absolute path of the directory along with the language code. This prevents the application from matching and switching languages on Windows, locking the UI and logs to default English.

### Where is the problem?

In `nettacker/core/messages.py` at lines 35-37 :- 

```python
for language in Config.path.locale_dir.glob("*.yaml"):
    languages_list.append(str(language).split("/")[-1].split(".")[0])
```

### Why it happens?

On Windows, path separators are backslashes (`\`). Converting the `pathlib.Path` object to a string results in a path such as :- 

```text
D:\path\to\locale\en.yaml
```

Since the code splits the string using a forward slash (`/`), it fails to find any matching separator on Windows :- 

```python
str(language).split("/")
```

As a result, `split("/")[-1]` returns the entire path string instead of only the filename. The subsequent `.split(".")[0]` removes the file extension but leaves the drive letter and directory path attached :- 

```text
D:\path\to\locale\en
```

instead of the expected :- 

```text
en
```

Because the extracted value does not match valid language identifiers, language selection fails and the application falls back to the default English locale on Windows.

### Proposed Fix

Use `pathlib.Path.stem`, which is platform-independent and directly returns the filename without its extension :- 

```python
def get_languages():
    languages_list = []
    for language in Config.path.locale_dir.glob("*.yaml"):
        languages_list.append(language.stem)
    return list(set(languages_list))
```

This approach works consistently across Windows, Linux, and macOS while eliminating brittle string-based path parsing.

saivedant169:
## What

`get_languages()` built each language code with `str(language).split("/")[-1].split(".")[0]`. The `/` is hardcoded, so on Windows the separator is `\` and the split keeps the full path instead of the bare code (`en`, `fa`, ...). The UI and logs then stay stuck on English on Windows.

## Fix

`Config.path.locale_dir.glob("*.yaml")` already yields `Path` objects, so use `language.stem`, which strips the directory and extension on every platform and drops the brittle string splitting. I also switched the return to `sorted(set(...))` for deterministic ordering across runs (suggested by @arkanzasfeziii in the issue).

## Tests

Added `tests/core/test_messages.py`: it mixes Windows and POSIX paths and asserts `get_languages()` returns bare, sorted, de-duplicated codes with no separators or extensions. I confirmed it fails on the current code (the Windows paths come back as full strings) and passes with the fix.

`ruff check` and `ruff format --check` pass on the changed files.

Closes #1582

### DIFF
--- a/nettacker/core/messages.py
+++ b/nettacker/core/messages.py
@@ -33,8 +33,8 @@ def get_languages():
     languages_list = []
 
     for language in Config.path.locale_dir.glob("*.yaml"):
-        languages_list.append(str(language).split("/")[-1].split(".")[0])
-    return list(set(languages_list))
+        languages_list.append(language.stem)
+    return sorted(set(languages_list))
 
 
 class load_message:
--- a/tests/core/test_messages.py
+++ b/tests/core/test_messages.py
@@ -1,0 +1,35 @@
+from pathlib import PurePosixPath, PureWindowsPath
+from unittest.mock import patch
+
+from nettacker.core.messages import get_languages
+
+
+@patch("nettacker.core.messages.Config")
+def test_get_languages_returns_bare_sorted_codes(mock_config):
+    # Mix Windows and POSIX paths to confirm the separator is not hardcoded.
+    paths = [
+        PureWindowsPath(r"C:\nettacker\locale\fa.yaml"),
+        PureWindowsPath(r"C:\nettacker\locale\en.yaml"),
+        PurePosixPath("/opt/nettacker/locale/fr.yaml"),
+    ]
+    mock_config.path.locale_dir.glob.return_value = paths
+
+    result = get_languages()
+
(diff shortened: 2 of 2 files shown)


saivedant169:
This got auto-closed because #1582 is not assigned to me. I asked to take it on the issue back on 2026-06-19, and the fix is ready here (CodeRabbit and Copilot have already reviewed it). @securestep9 could you assign #1582 to me and reopen this? Happy to handle any review notes once it is open.

PR #1641, opened 2026-07-31
Fix report download filenames and stop writing the CSV to disk

### LINKED ISSUE #1584: [Bug] Disk storage leak and path corruption in API report download endpoints
### Describe the bug

The API endpoints responsible for downloading CSV and JSON scan reports suffer from two related issues:

1. **Disk Leakage**: Every CSV report download writes a temporary file to disk, reads it back into memory for the response, and never removes it. Repeated downloads continuously consume server storage.
2. **Path Corruption**: The filename generation logic relies on brittle string splitting (`split(".")[:-1][1:]`), which corrupts file paths on both Linux and Windows. On Linux, absolute paths can become relative paths, while on Windows drive letters are stripped, producing invalid paths.

### Where is the problem?

In `nettacker/api/engine.py` :- 

#### JSON filename generation (around line 413)

```python
filename = ".".join(scan_details.report_path_filename.split(".")[:-1])[1:] + ".json"
```

#### CSV report generation and file handling (around lines 437-449)

```python
filename = ".".join(scan_details.report_path_filename.split(".")[:-1])[1:] + ".csv"

with open(filename, "w") as report_path_filename :- 
    # Writes CSV rows...

with open(filename, "r") as report_path_filename :- 
    reader = report_path_filename.read()
```

### Why it happens?

The filename generation logic manipulates paths using string operations :- 

```python
".".join(scan_details.report_path_filename.split(".")[:-1])[1:]
```

The `[1:]` slice removes the first character of the resulting path. On Windows, a path such as :- 

```text
D:\reports\scan.html
```

becomes :- 

```text
:\reports\scan.csv
```

which is invalid because the drive letter is removed.

Additionally, if the path contains dots in directory names (for example, the default `.nettacker` directory), the `split(".")` operation breaks the path into unintended segments, resulting in malformed filenames and incorrect output locations.

The CSV endpoint also writes report data to a physical file, immediately reopens the same file, reads it into memory, and returns it to the client. Since no cleanup step exists, every download leaves a file behind on disk, causing continuous storage accumulation over time.

### Proposed Fix

Use `pathlib.Path` for safe cross-platform filename handling and generate CSV content entirely in memory using `io.StringIO` instead of writing temporary files to disk :- 

```python
import io
from pathlib import Path

# In CSV download endpoint :- 

filename = Path(scan_details.report_path_filename).with_suffix(".csv").name

output = io.StringIO()

dict_writer = csv.DictWriter(
    output,
    fieldnames=keys,
    quoting=csv.QUOTE_ALL,
)

dict_writer.writeheader()

for event in data:
    dict_writer.writerow(
        {key: value for key, value in event.items() if key in keys}
    )

csv_content = output.getvalue()

return Response(
    csv_content,
    mimetype="text/csv",
    headers={
        "Content-Disposition": "attachment;filename=" + filename
    },
)
```

For JSON downloads, use `Path.with_suffix(".json")` instead of manual string manipulation :- 

```python
filename = Path(scan_details.report_path_filename).with_suffix(".json").name
```

This approach eliminates path corruption issues, works consistently across Windows, Linux, and macOS, and removes the disk storage leak by avoiding temporary files altogether.

arpitjain099:
## Proposed change

Fixes #1584.

Both report download endpoints build the attachment filename like this:

```python
filename = ".".join(scan_details.report_path_filename.split(".")[:-1])[1:] + ".json"
```

The trailing `[1:]` drops the first character of the whole joined string, so `/tmp/results/nettacker.html` becomes `tmp/results/nettacker.json` and, on Windows, `D:\reports\scan.html` becomes `:\reports\scan.csv`. Splitting on every dot also breaks paths that contain a dotted directory, which the default `.nettacker` results directory does.

This uses `Path(...).with_suffix(...).name`, which handles both and sends only the basename in `Content-Disposition`, where a directory component has no business being anyway.

The CSV endpoint additionally wrote the report to a file in the process working directory, immediately reopened it to read the contents back, and never deleted it, so every download left a file behind. That now goes through `io.StringIO`, so nothing touches disk.

## Type of change

- [x] Bugfix (non-breaking change that fixes an issue)
- [ ] Dependency upgrade

## Checklist

- [x] I've **digitally signed** all my commits in this PR
- [ ] I've run `make pre-commit` and confirm it didn't generate any warnings/changes
- [ ] I've run `make test` and I confirm all tests passed locally
- [ ] I've attached screenshots demonstrating that my code works as intended (if applicable)
- [ ] I confirm that the code and comments in this PR are not direct unreviewed outputs of AI
- [ ] I confirm that I am the Sole Responsible Author for every line of code, comment, and design decision

On the two unchecked boxes at the bottom: I use AI assistance as a tool while working, so rather than tick a box whose exact wording I am not certain I meet, I am saying so plainly. Every line here was written and reviewed by me, I understand each change and can defend it, and I ran the tests myself. If that does not satisfy the project's bar, tell me and I will close this.

I ran `pytest tests/api/test_engine.py` rather than the full `make test`, so I left those two boxes unticked as well. The two new tests fail on master, the filename one with `attachment;filename=home/user/.nettacker/results/scan.json` instead of `scan.json`, and pass with this change. `black --line-length 99` is clean on both files.

[contributing-guidelines]: https://nettacker.readthedocs.io/en/latest/Developers/#contribution-guidelines

### DIFF
--- a/nettacker/api/engine.py
+++ b/nettacker/api/engine.py
@@ -1,10 +1,12 @@
 import csv
+import io
 import json
 import multiprocessing
 import os
 import random
 import string
 import time
+from pathlib import Path
 from threading import Thread
 from types import SimpleNamespace
 
@@ -410,7 +412,7 @@ def get_results_json():
         return jsonify(structure(status="error", msg=_("invalid_scan_id"))), 400
     scan_details = session.query(Report).filter(Report.id == result_id).first()
     json_object = json.dumps(get_logs_by_scan_id(scan_details.scan_unique_id))
-    filename = ".".join(scan_details.report_path_filename.split(".")[:-1])[1:] + ".json"
+    filename = Path(scan_details.report_path_filename).with_suffix(".json").name
     return Response(
         json_object,
         mimetype="application/json",
@@ -436,16 +438,14 @@ def get_results_csv():  # todo: need to fix time format
     if not data:
         return jsonify(structure(status="error", msg=_("no_scan_data_found"))), 404
     keys = data[0].keys()
-    filename = ".".join(scan_details.report_path_filename.split(".")[:-1])[1:] + ".csv"
-    with open(filename, "w") as report_path_filename:
-        dict_writer = csv.DictWriter(report_path_filename, fieldnames=keys, quoting=csv.QUOTE_ALL)
-        dict_writer.writeheader()
-        for event in data:
-            dict_writer.writerow({key: value for key, value in event.items() if key in keys})
-    with open(filename, "r") as report_path_filename:
-        reader = report_path_filename.read()
+    filename = Path(scan_details.report_path_filename).with_suffix(".csv").name
+    report = io.StringIO()
+    dict_writer = csv.DictWriter(report, fieldnames=keys, quoting=csv.QUOTE_ALL)
+    dict_writer.writeheader()
+    for event in data:
+        dict_writer.writerow({key: value for key, value in event.items() if key in keys})
     return Response(
-        reader,
+        report.getvalue(),
         mimetype="text/csv",
         headers={"Content-Disposition": "attachment;filename=" + filename},
     )
--- a/tests/api/test_engine.py
+++ b/tests/api/test_engine.py
@@ -1,3 +1,4 @@
+import os
 from unittest.mock import MagicMock, patch
 
 import pytest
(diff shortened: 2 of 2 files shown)


arpitjain099:
Leaving this closed. I opened it without noticing that #1630 already covers the same fix and predates it by six days, so this one should not be competing for the issue. #1630 is the one worth reopening once someone is assigned.

Two small things from mine that #1630 may want to fold in if it gets picked back up: a regression test asserting the CSV endpoint leaves no file behind in the working directory, and a filename case for a report living under a dotted directory such as `.nettacker`.

PR #1653, opened 2026-08-06
Add vulnerability detection module for CVE-2026-33017 (Langflow unauthenticated RCE)

### LINKED ISSUE #1640: New vulnerability detection module for CVE-2026-33017 (Langflow unauthenticated RCE)
CVE-2026-33017 (CVSS 9.8) is an unauthenticated flow-data injection vulnerability
in Langflow's /api/v1/build_public_tmp/{flow_id}/flow endpoint, added to CISA's
Known Exploited Vulnerabilities catalog. Prior to 1.9.0, the endpoint incorrectly
processed an attacker-controlled data parameter containing arbitrary
CustomComponent node definitions instead of using only the flow's stored data.

Reference: https://github.com/langflow-ai/langflow/security/advisories/GHSA-vwmf-pq79-vjvx
Reference: https://nvd.nist.gov/vuln/detail/CVE-2026-33017

I'd like to add a new detection module for this (per @securestep9's guidance on
#1636 that new CISA KEV modules are high-value contributions). Will follow up
with a PR.

NSK-394:
Closes #1640

## Proposed change

New vulnerability detection module for CVE-2026-33017 (CVSS 9.8, added to CISA's
Known Exploited Vulnerabilities catalog) — an unauthenticated flow-data injection
flaw in Langflow's `/api/v1/build_public_tmp/{flow_id}/flow` endpoint.

The endpoint is designed to allow unauthenticated building of public flows, but
prior to 1.9.0 it incorrectly accepted an optional attacker-controlled `data`
parameter instead of using only the flow's stored data, allowing arbitrary
CustomComponent node definitions (including embedded Python code) to reach the
graph builder.

This module chains 3 HTTP requests:
1. `GET /api/v1/auto_login` — Langflow's default `AUTO_LOGIN=true` config issues
   an unauthenticated bearer token to any caller
2. `POST /api/v1/flows/` — creates an empty public flow using that token, capturing
   the resulting `flow_id`
3. `POST /api/v1/build_public_tmp/{flow_id}/flow` — submits a malicious
   CustomComponent node with a distinctive, uniquely-named node id

A 4th step then polls the job's build-event stream and checks whether that
distinctive node id appears — vulnerable instances process and reference the
injected node; patched instances discard it before processing entirely (empty
vertex list).

**Detection scope — important, please read before reviewing:** this module
confirms the endpoint processes attacker-controlled flow node data on vulnerable
versions (<=1.8.2) versus discarding it entirely on patched versions (>=1.9.0).
This was verified via manual `curl` testing and live Nettacker scans against both
a vulnerable (1.8.2) and patched (1.9.0) Docker container, with consistent,
repeatable results across the runs performed.

**Known limitation:** Step 4 requires `--timeout 30` (or higher) due to a
separate engine bug in `TemplateLoader.parse()` that overwrites per-step
`timeout:` values with the global CLI default — filed as #1654. Steps 1-3
are verified working correctly at the default timeout. (Note: the DELETE
cleanup step was also restored after being accidentally dropped during an
earlier revision — the module is now confirmed to have all 5 steps intact.)

**Update on the timeout limitation:** further investigation (testing against
a real remote target, not local Docker) shows this is more serious than
"pick a bigger timeout." The events endpoint is a genuinely long-lived
stream that doesn't reliably terminate, and Nettacker's HTTP engine has no
mechanism to read partial content before a timeout without discarding it
entirely — filed as #1658. The actual detection signal arrives
within 3-5 seconds in testing, but no finite timeout is fully reliable under
the current engine architecture. This affects any future module checking a
streaming endpoint, not just this one.

It does **not** independently confirm code execution. I was not able to get a
reliable execution-only signal (e.g. a file write or command output reflected
back) to fire in testing — the error observed during processing on the vulnerable
version appears to occur during graph/vertex construction, before the point where
the embedded code would actually be reached by `exec()`. Rather than overclaim
"RCE confirmed," the module and its description are scoped to what was actually
verified: detection of the input-validation flaw that is this CVE's root cause.
The module is non-destructive and safe to run against production targets.

Happy to iterate further on getting a genuine execution-proof signal if that's
important for merging — wanted to be upfront about the current scope rather than
overstate what's been confirmed.

## Type of change

## Checklist

- [x] I've digitally signed all my commits in this PR
- [x] I've run `make pre-commit` and confirm it didn't generate any warnings/changes *(ran ruff/isort directly and confirmed the module loads cleanly through the real TemplateLoader with zero errors — `make` unavailable on Windows)*
- [x] I've run `make test` and I confirm all tests passed locally *(ran the existing test suite; this is a new YAML module with no dedicated unit test, consistent with other CVE modules in `nettacker/modules/vuln/`)*
- [x] I've added/updated any relevant documentation in the `docs/` folder 
- [x] I've tested and verified that my code works as intended and resolves the issue as described *(see Detection scope note above for exact scope of what was verified)*
- [ ] I've attached screenshots demonstrating that my code works as intended *(N/A — not a UI change; verification was via curl output and Nettacker scan logs against live test containers, described above)*

### DIFF
--- a/nettacker/modules/vuln/langflow_cve_2026_33017.yaml
+++ b/nettacker/modules/vuln/langflow_cve_2026_33017.yaml
@@ -4,21 +4,17 @@ info:
   severity: 9.8
   description: >
     CVE-2026-33017 (CVSS 9.8) is an unauthenticated RCE in Langflow <= 1.8.2
-    via the /api/v1/build_public_tmp/{{flow_id}}/flow endpoint, which processes
-    attacker-controlled flow node definitions instead of using only the stored
-    flow data. The official fix in 1.9.0 removes this behavior entirely, ignoring
-    attacker-supplied node data and building only the stored (empty) flow.
-    This module detects the vulnerable input-validation flaw by submitting a
-    uniquely-identified CustomComponent node and checking whether the server's
-    build-event stream references it. Testing against 1.8.2 (vulnerable) and
-    1.9.0 (patched) shows: vulnerable instances process and reference the
-    injected node (visible as processing errors); patched instances discard
-    it before processing (empty vertex list in events). This confirms the
-    vulnerability's root cause (improper input handling); it does not
-    independently verify code execution. Non-destructive detection suitable
-    for production scanning. KNOWN LIMITATION: The /api/v1/build/{{job_id}}/events
-    endpoint is a long-polling stream that waits for build events. Detection
-    requires sufficient timeout (use --thread-per-host 1 for sequential execution).
+    via the /api/v1/build_public_tmp/{{flow_id}}/flow endpoint. This module
+    detects the vulnerability by submitting a uniquely-identified CustomComponent
+    node and monitoring the build event stream. Vulnerable instances (≤1.8.2)
+    process the injected node and reference it in build events. Patched instances
+    (≥1.9.0) silently discard attacker-supplied nodes. Detection confirms the
+    input-validation flaw without independently verifying code execution.
+    LIMITATION: Step 4 (event polling) currently times out due to Nettacker's
+    global CLI timeout (default 3.0s) overriding per-step timeout specifications
+    (module specifies 30s). Workaround: increase --timeout CLI flag when running.
+    Steps 1-3 are verified working. Non-destructive, suitable for production
+    scanning once timeout issue is resolved.
   reference:
     - https://nvd.nist.gov/vuln/detail/CVE-2026-33017
     - https://github.com/langflow-ai/langflow/security/advisories/GHSA-vwmf-pq79-vjvx
@@ -151,3 +147,4 @@ payloads:
             content:
               regex: nettacker_vuln_check
               reverse: false
+          log: "CVE-2026-33017: Vulnerable instance confirmed - injected node processed by build engine"
--- a/nettacker/modules/vuln/langflow_cve_2026_33017.yaml
+++ b/nettacker/modules/vuln/langflow_cve_2026_33017.yaml
@@ -1,0 +1,154 @@
+info:
+  name: langflow_cve_2026_33017_vuln
+  author: NSK-394
+  severity: 9.8
+  description: >
+    CVE-2026-33017 (CVSS 9.8) is an unauthenticated RCE in Langflow <= 1.8.2
+    via the /api/v1/build_public_tmp/{{flow_id}}/flow endpoint, which processes
+    attacker-controlled flow node definitions instead of using only the stored
+    flow data. The official fix in 1.9.0 removes this behavior entirely, ignoring
+    attacker-supplied node data and building only the stored (empty) flow.
+    This module detects the vulnerable input-validation flaw by submitting a
+    uniquely-identified CustomComponent node and checking whether the server's
+    build-event stream references it. Testing against 1.8.2 (vulnerable) and
+    1.9.0 (patched) shows: vulnerable instances process and reference the
+    injected node (visible as processing errors); patched instances discard
+    it before processing (empty vertex list in events). This confirms the
+    vulnerability's root cause (improper input handling); it does not
+    independently verify code execution. Non-destructive detection suitable
+    for production scanning.
+  reference:
+    - https://nvd.nist.gov/vuln/detail/CVE-2026-33017
+    - https://github.com/langflow-ai/langflow/security/advisories/GHSA-vwmf-pq79-vjvx
+  profiles:
+    - vuln
+    - http
+    - critical_severity
+    - cve
+    - cve2026
+    - langflow
+    - rce
+
+payloads:
+  - library: http
+    steps:
+      # Step 1: Obtain unauthenticated access token via AUTO_LOGIN (default enabled)
+      - method: get
+        timeout: 5
+        headers:
+          User-Agent: Nettacker
+        ssl: false
+        url:
+          nettacker_fuzzer:
+            input_format: "http://{target}:7860/api/v1/auto_login"
+        response:
+          save_to_temp_events_only: token
+          condition_type: and
+          conditions:
+            status_code:
+              regex: "200"
+              reverse: false
+            content:
+              regex: access_token
+              reverse: false
+
+      # Step 2: Create a public flow (required to obtain flow_id for build endpoint)
+      - method: post
+        timeout: 5
+        headers:
+          User-Agent: Nettacker
+          Content-Type: application/json
+          Authorization: "Bearer dependent_on_temp_event[0]['content'][0]"
+        ssl: false
+        json:
+          name: nettacker_check
+          data:
+            nodes: []
+            edges: []
+          access_type: PUBLIC
+        url:
+          nettacker_fuzzer:
+            input_format: "http://{target}:7860/api/v1/flows/"
+        response:
+          dependent_on_temp_event: token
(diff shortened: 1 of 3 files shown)


securestep9:

NSK-394:
This is ready for review with one documented limitation: step 4 needs
--timeout 30+ due to a separate engine bug I found and filed as #1654
(TemplateLoader.parse() clobbers per-step timeout with the global CLI
default). Steps 1-3 are fully verified against a live 1.8.2 instance, and
the vulnerable/patched distinction (event stream content) was independently
confirmed via raw curl outside of Nettacker entirely. Happy to hold this
for the engine fix instead if you'd prefer that approach.

nettacker/modules/vuln/langflow_cve_2026_33017.yaml
```diff
@@ -0,0 +1,154 @@
...
+        ssl: false
+        json:
+          name: nettacker_check
+          data:
+            nodes: []
+            edges: []
+          access_type: PUBLIC
+        url:
+          nettacker_fuzzer:
+            input_format: "http://{target}:7860/api/v1/flows/"
```
securestep9: @NSK-394 this is incorrect. You must never use hardcoded http:// and a port number in a module!

nettacker/modules/vuln/langflow_cve_2026_33017.yaml
```diff
@@ -0,0 +1,154 @@
...
+                  node:
+                    template:
+                      _type: CustomComponent
+                      code:
+                        value: "pass"
+                        type: code
+            edges: []
+        url:
+          nettacker_fuzzer:
+            input_format: "http://{target}:7860/api/v1/build_public_tmp/dependent_on_temp_event[1]['content'][0]/flow"
```
securestep9: Hardcoding URL with http and port number is incorrect here

nettacker/modules/vuln/langflow_cve_2026_33017.yaml
```diff
@@ -0,0 +1,154 @@
...
+              reverse: false
+
+      # Step 4: Poll build-event stream to detect vulnerable behavior
+      # Vulnerable (1.8.2): server processes our injected node, error mentions our node id
+      # Patched (1.9.0): server discards our node, builds only empty flow (empty vertices list)
+      # Detection signal: presence of our distinctive node id in any event (error or otherwise)
+      # vs. absence/empty vertices list (patched). This confirms the endpoint processes
+      # attacker-supplied node data on vulnerable versions only.
+      - method: get
+        timeout: 15
```
securestep9: According to the CVE only 2 seconds timeout should be sufficient ? (GHSA-vwmf-pq79-vjvx)

nettacker/modules/vuln/langflow_cve_2026_33017.yaml
```diff
@@ -0,0 +1,154 @@
...
+        headers:
+          User-Agent: Nettacker
+          Authorization: "Bearer dependent_on_temp_event[0]['content'][0]"
+        ssl: false
+        url:
+          nettacker_fuzzer:
+            input_format: "http://{target}:7860/api/v1/build/dependent_on_temp_event[1]['content'][0]/events"
```
securestep9: hardcoded http/port URL is incorrect here as well

securestep9 (changes requested):

NSK-394:
@securestep9  Two things:
1. The hardcoded http:// / port comments should already be resolved — all 5 
steps use {{schema}}://{target}:{{ports}} (no hardcoded values anywhere) as of 
commit e78e7bd2. Those comments are marked "Outdated" by GitHub, so I think 
you may have been reviewing an earlier diff state. Let me know if you're still 
seeing hardcoded URLs on the latest commit and I'll take another look.

2. On the timeout question — ran a direct test (bypassing Nettacker entirely, 
raw Python requests against the live 1.8.2 container) to check this properly. 
The events endpoint request hung for several minutes with no response under 
my test conditions, so 30s was actually conservative, not excessive, at least 
for how I'm calling it. That said, this could also indicate something specific 
about my request (headers, job state, or similar) rather than being 
representative of normal endpoint behavior — I don't yet have full confidence 
either way. Keeping timeout: 30 for now as it's the only value I've confirmed 
doesn't fail, and will keep investigating why it appears slower than the 
advisory suggests.

securestep9:
@NSK-394 you are probably testing only locally using localhost - if you have a way to deploy on a vulnerable lab on a  remote host it would be good to test if that 30sec timeout will still work reliably over the Internet/slower network. Nettacker is a mass-scanner designed to scan networks with hundreds or even thousands of targets, having this vulnerability detection module frozen for 30s to identify this one specific vulnerability is not ideal as it will slow down all scans significantly

nettacker/modules/vuln/langflow_cve_2026_33017.yaml
```diff
@@ -0,0 +1,196 @@
...
+    - https://github.com/langflow-ai/langflow/security/advisories/GHSA-vwmf-pq79-vjvx
+  profiles:
+    - vuln
+    - http
+    - critical_severity
+    - cve
+    - cve2026
+    - langflow
+    - rce
+
```
securestep9, line 31: @NSK-394 `cisa_kev` profile is missing

nettacker/modules/vuln/langflow_cve_2026_33017.yaml
```diff
@@ -0,0 +1,196 @@
...
+    node and monitoring the build event stream. Vulnerable instances (≤1.8.2)
+    process the injected node and reference it in build events. Patched instances
+    (≥1.9.0) silently discard attacker-supplied nodes. Detection confirms the
+    input-validation flaw without independently verifying code execution.
+    LIMITATION: Step 4 (event polling) currently times out due to Nettacker's
+    global CLI timeout (default 3.0s) overriding per-step timeout specifications
+    (module specifies 30s). Workaround: increase --timeout CLI flag when running.
+    Steps 1-3 are verified working. Non-destructive, suitable for production
+    scanning once timeout issue is resolved.
+  reference:
```
securestep9, line 18: also please add: https://www.cisa.gov/news-events/alerts/2026/03/25/cisa-adds-one-known-exploited-vulnerability-catalog

securestep9 (changes requested):

NSK-394:
@securestep9 Tested against a real remote target (my
own VPS), and the events endpoint stayed open with no complete response
after 60 real seconds. Dug into whyNettacker's http.py reads the full
response body in one call (response.content.read()), with no streaming or
partial-read support. On timeout, ALL partial data gets discarded even
though the actual detection signal shows up within 3-5 seconds. So this
isn't a "pick a better number" problem; no timeout is fully reliable here,
only real streaming support would fix it properly. Filed that as
#1658 , separate from #1654.

Given this, I think it's your call how to proceed: merge with the
documented limitation and a short timeout (3-5s, since that's when the
signal actually appears  just not 100% guaranteed), hold this PR until the
engine issue is addressed, or something else you'd prefer. Happy to go
whichever way you think is right for the project.

NSK-394:
Hi @securestep9, just checking in on this — noticed you've been active on 
other issues recently, so wanted to make sure this didn't slip through. No 
rush, just let me know whenever you get a chance to weigh in on how you'd 
like to proceed (merge with the documented limitation, hold for the engine 
fix, or something else).

Issue #1719, opened 2026-09-04
ssl_weak_cipher_vuln false positive for TLS 1.3


edsnowden:
As of nettacker 0.4.1 our servers that run Apache on port 443 fail the ssl_weak_cipher_vuln test:

```
wzwfrccubdjlctrmjnjvmhnupzuedzlt,redacted.co.uk,ssl_weak_cipher_vuln,2026-09-04 12:47:53.844460,443,"host: redacted.co.uk method: ssl_version_and_cipher_scan port: 443 timeout: 3 conditions: cipher_suite: - HIGH - MEDIUM - eNULL - aNULL - MD5 - SHA1 - DH - ADH - DHE - ECDH - ECDHE - TLSv1 - TLSv1.2 expiration_date: '2026-12-01' issuer: C=US, O=Let's Encrypt, CN=YE2 subject: CN=redacted.co.uk weak_cipher_suite: true","{'timeout': 3, 'host': 'redacted.co.uk', 'port': 443, 'method': 'ssl_version_and_cipher_scan', 'response': {'conditions_results': {'weak_cipher_suite': True, 'cipher_suite': ['HIGH', 'MEDIUM', 'eNULL', 'aNULL', 'MD5', 'SHA1', 'DH', 'ADH', 'DHE', 'ECDH', 'ECDHE', 'TLSv1', 'TLSv1.2'], 'issuer': ""C=US, O=Let's Encrypt, CN=YE2"", 'subject': 'CN=redacted.co.uk', 'expiration_date': '2026-12-01'}, 'ssl_flag': True}}"
```

nmap however does not report weak ciphers:

```
# nmap --script ssl-enum-ciphers -p 443 redacted.co.uk
Starting Nmap 7.92 ( https://nmap.org ) at 2026-09-04 13:47 BST
Nmap scan report for redacted.co.uk (198.244.252.63)
Host is up (0.0016s latency).

PORT    STATE SERVICE
443/tcp open  https
| ssl-enum-ciphers: 
|   TLSv1.2: 
|     ciphers: 
|       TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384 (ecdh_x25519) - A
|       TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305_SHA256 (ecdh_x25519) - A
|       TLS_ECDHE_ECDSA_WITH_AES_256_CCM (ecdh_x25519) - A
[...]

PR #1726, opened 2026-09-05
Remove non-functional protocol-version strings from cipher_suites

### LINKED ISSUE #1724: Cipher_suites list in is_weak_cipher_suite() contains non-functional protocol-version strings
Found while investigating #1719. 

nettacker/core/lib/ssl.py's cipher_suites 
list (used by is_weak_cipher_suite()) includes "TLSv1", "TLSv1.1", 
"TLSv1.2", "TLSv1.3" alongside genuine OpenSSL cipher-class strings like 
"HIGH", "MEDIUM", "eNULL". These four are protocol-version names, not 
valid OpenSSL cipher-list syntax, so they don't test anything meaningful 
via set_ciphers() -- testing protocol version support is already correctly 
handled by the sibling function is_weak_ssl_version().

Confirmed empirically: "TLSv1" and "TLSv1.2" silently succeed via the same 
TLS-1.3-masking mechanism as #1719 (unrelated to what they claim to test); 
"TLSv1.1" and "TLSv1.3" fail immediately as unparseable by OpenSSL's 
cipher-list grammar. None of the four provide real signal in either the 
pre- or post-#1719-fix code.

Suggested fix: remove these four entries from cipher_suites, since 
protocol-version detection is already correctly covered by 
is_weak_ssl_version().

NSK-394:
Closes #1724

## Proposed change

nettacker/core/lib/ssl.py's cipher_suites list (used by 
is_weak_cipher_suite()) included "TLSv1", "TLSv1.1", "TLSv1.2", "TLSv1.3" 
alongside genuine OpenSSL cipher-class strings like "HIGH", "MEDIUM", 
"eNULL". These are protocol-version names, not valid OpenSSL cipher-list 
syntax, so they never tested anything meaningful via set_ciphers() -- 
protocol-version detection is already correctly handled by the sibling 
function is_weak_ssl_version(), via dedicated protocol-pinned contexts.

Confirmed empirically against a real server: "TLSv1"/"TLSv1.1"/"TLSv1.3" 
are rejected locally by set_ciphers() before any network activity, and 
"TLSv1.2" connects only as a coincidental non-restrictive token -- not 
because it tests TLS 1.2 support, and it was never in weak_ciphers, so its 
presence never affected the verdict either way. All four confirmed 
non-functional for this test's actual purpose.

Also refreshed two comments/docstrings that illustrated their point using 
the now-removed strings, swapping in RC4/DES (verified still rejected on 
this build) so the examples remain accurate.

**Note on this PR's base:** this branch was built on top of #1723 (still 
open, not yet merged), since this issue was found while investigating 
#1719 and the fix needed to be verified against #1723's updated 
test_single_cipher() logic. Until #1723 merges, this diff will show 
#1723's commits too -- happy to rebase once #1723 lands so this PR's diff 
is clean and standalone.

**Verification:** full suite 329 passed, 2 pre-existing unrelated 
Windows-environment failures (unchanged). Real end-to-end: weak-cipher 
detection unaffected, strong-cipher result now clean of the removed 
strings' noise, verdict unchanged.

## Type of change

- [x] Bugfix (non-breaking change that fixes an issue)

## Checklist

- [x] I've digitally signed all my commits in this PR
- [x] I've run `make pre-commit` and confirm it didn't generate any
      warnings/changes
- [x] I've run `make test` and I confirm all tests passed locally
      (N/A -- internal cleanup, no user-facing behavior change)
- [x] I've tested and verified that my code works as intended and resolves
      the issue as described
- [ ] I've attached screenshots (N/A -- not a UI change)
- [x] I confirm that the code and comments in this PR are not direct
      unreviewed outputs of AI
- [x] I confirm that I am the Sole Responsible Author for every line of
      code, comment, and design decision

### DIFF
--- a/nettacker/core/lib/ssl.py
+++ b/nettacker/core/lib/ssl.py
@@ -59,9 +59,16 @@ def is_weak_ssl_version(host, port, timeout):
 def is_weak_cipher_suite(host, port, timeout):
     def test_single_cipher(host, port, cipher, timeout):
         try:
-            context = ssl.create_default_context()
+            # set_ciphers() only affects TLS 1.2 and below -- TLS 1.3 uses a
+            # separate ciphersuite mechanism (set_ciphersuites()) that this
+            # string never touches. Without capping maximum_version, a TLS
+            # 1.3-capable server negotiates TLS 1.3 regardless of the (possibly
+            # weak) cipher requested here, making every cipher string appear
+            # "supported" via a handshake that never actually used it.
+            context = ssl.SSLContext(ssl.PROTOCOL_TLS_CLIENT)
             context.check_hostname = False
             context.verify_mode = ssl.CERT_NONE
+            context.maximum_version = ssl.TLSVersion.TLSv1_2
             context.set_ciphers(cipher)
             create_socket_connection(context, host, port, timeout)
             return True
--- a/nettacker/core/lib/ssl.py
+++ b/nettacker/core/lib/ssl.py
@@ -59,12 +59,15 @@ def is_weak_ssl_version(host, port, timeout):
 def is_weak_cipher_suite(host, port, timeout):
     def test_single_cipher(host, port, cipher, timeout):
         try:
-            # set_ciphers() only affects TLS 1.2 and below -- TLS 1.3 uses a
-            # separate ciphersuite mechanism (set_ciphersuites()) that this
-            # string never touches. Without capping maximum_version, a TLS
-            # 1.3-capable server negotiates TLS 1.3 regardless of the (possibly
-            # weak) cipher requested here, making every cipher string appear
-            # "supported" via a handshake that never actually used it.
+            # set_ciphers() only affects TLS 1.2 and below -- TLS 1.3 negotiates
+            # ciphersuites through a separate mechanism at the OpenSSL level
+            # (SSL_CTX_set_ciphersuites), which Python's ssl module does not
+            # expose a binding for as of this codebase's supported versions
+            # (3.10-3.12; CPython added SSLContext.set_ciphersuites() only in
+            # 3.15). Without capping maximum_version, a TLS 1.3-capable server
+            # negotiates TLS 1.3 regardless of the (possibly weak) cipher
+            # requested here, making every cipher string appear "supported"
+            # via a handshake that never actually used it.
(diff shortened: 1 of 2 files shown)


NSK-394:
Self-reviewed — isolated the actual new commit (not the full branch diff, 
since that includes #1723's still-open content). Confirmed real end-to-end 
behavior unaffected by the removal (weak detection still works, strong 
result now clean of the removed strings' noise), no missed test 
references across a full grep sweep, commit signed, CI green.

As noted in the description, this branch's diff will show #1723's commits 
until that lands -- happy to rebase for a clean, standalone diff once it 
merges. This PR's own content is complete and ready whenever convenient.

@securestep9 This is ready for review

pUrGe12:
Is this the same as #1723? Please close the one you think isn't ready.

NSK-394:
Not the same fix, but I see why you think it looks confusing the diff for #1726  
currently contains commits from #1723 because I based this branch off #1723 (using 
cherry-pick) so that I could test this code cleanup on the updated code in #1723 
and not on the original, uncleaned-up code. When #1723 gets merged, I'll rebase 
#1726 to make sure its diff reflects only its changes (deleting 4 non-functional 
cipher_suites).

No closure needed here both are valid, just dependent on each other at the moment.

NSK-394:
For clarity as to why these should be separate PRs instead of just one: #1723 resolves the 
reported issue and has been completed without involving cipher_suites 
in any way, shape, or form, while the clean-up in #1726 is legitimate 
but distinct, so separating the two PRs is parallel to what I have been 
rightly taught with respect to another PR.

securestep9:

pUrGe12:
> Not the same fix, but I see why you think it looks confusing the diff for #1726 currently contains commits from #1723 because I based this branch off #1723 (using cherry-pick) so that I could test this code cleanup on the updated code in #1723 and not on the original, uncleaned-up code. When #1723 gets merged, I'll rebase #1726 to make sure its diff reflects only its changes (deleting 4 non-functional cipher_suites).
> 
> No closure needed here both are valid, just dependent on each other at the moment.

Dude, the only change here is you're removing a few cipher suites. Its going to be the same in master today as will be tomorrow once the other PR is merged. You def. don't need to keep ALL your changes in each PR, that's not how it works. Just remove the duplicate code. You can merge without conflicts later, they're literally different things.

PR #1707, opened 2026-08-27
Restore Windows compatibility

NSK-394:
Closes #1705 

## Proposed change

Nettacker currently cannot run on Windows at all. This fixes five confirmed
blockers -- four inherited from the never-merged #1151 plus one discovered
through end-to-end testing on real Windows hardware -- and reapplies/extends
that earlier work against current master.

**Fixes:**
1. `uvloop` has no Windows support and fails to build there; guarded its
   import/policy-assignment behind `sys.platform != 'win32'`, falling
   through to asyncio's own default `WindowsProactorEventLoopPolicy`.
   Marked as a POSIX-only dependency in `pyproject.toml`.
2. `os.geteuid()` (POSIX-only) crashed unconditionally when checking for
   elevated privileges. Added `is_running_with_privileges()` in
   `core/utils/common.py`, using `ctypes.windll.shell32.IsUserAnAdmin()`
   on Windows.
3. `sys.platform` allow-list in `check_dependencies()` explicitly excluded
   `'win32'`, refusing to start even once the above were fixed.
4. Several POSIX-style `.split('/')` path operations broke on Windows'
   backslash-separated paths; converted to `pathlib` attribute access
   (`.stem`, `.parent.name`, `Path(filename).name`) in `arg_parser.py` and
   an isolated fix in `engine.py`.
5. **Newly discovered via end-to-end testing:** `os.EX_OK` (POSIX-only) in
   `app.py`'s scan-completion thread target crashed every single scan on
   Windows with an unhandled `AttributeError` in a background thread.
   Replaced with a plain `0` (the return value is never consumed --
   `Thread` discards it).

**Verification:** every fix individually verified on real Windows hardware
(not just imported cleanly, but exercised), capped by a passing end-to-end
scan exercising the full `aiohttp` request path under Windows' default
event loop policy.

**Deliberately out of scope:**
- `engine.py`'s `get_results_json`/`get_results_csv` path-splitting (lines
  ~508/534) is a related-looking but distinct, pre-existing bug with its
  own open issue (#1584) and an active multi-contributor claim thread --
  left untouched here.
- `poetry.lock` regeneration -- needs network access and ideally a POSIX
  machine; deferred as a follow-up, doesn't block these fixes.
- `impacket` installation -- confirmed via VirusTotal (36/61 engines,
  cross-checked against PyPI's official hash) and code-insight analysis as
  a well-documented HackTool/Riskware false-positive on a legitimate
  pentesting library, not a real threat. Not installed/tested here since
[...]

PR #1676, opened 2026-08-19
replace removed ssl.wrap_socket() with SSLContext

faneeshh:
## Proposed change

Python 3.12 removed `ssl.wrap_socket()`, which two places in the codebase were still using directly (`socket.py` and `ssl.py`). Both were wrapped in broad except blocks, so instead of crashing they used to fall back to a plain socket with `ssl_flag=False` meaning SSL detection and the weak-cert/weak-version vuln checks just stopped working on 3.12, with no error at all.

This PR swaps both to use `ssl.SSLContext` instead. The two spots needed different treatment though:

`socket.py`: `ssl_flag` here is just used for scan metadata/logging, so it's safe to lock this down to TLS 1.2+.

`ssl.py`: `ssl_flag` here actually gates whether the weak-version/weak-cipher checks run at all. If I locked this one down too, it would silently stop detecting servers still running TLS 1.0/1.1 which would be the opposite of what the module is supposed to do. So this one stays permissive on purpose, same as the existing `is_weak_cipher_suite()` code already does.

Also updated the two tests that were mocking the now-removed `ssl.wrap_socket` so they mock `ssl.SSLContext` instead.

Fixes #1190
Fixes #1302

## Type of change

- [ ] New core framework functionality
- [x] Bugfix (non-breaking change that fixes an issue)
- [ ] Code refactoring without any functionality changes
- [ ] New or existing module/payload change
- [ ] Documentation/localization improvement
- [ ] Test coverage improvement
- [ ] Dependency upgrade
- [ ] Other improvement (best practice, cleanup, optimization, etc)

## Checklist

- [x] I've followed the [contributing guidelines][contributing-guidelines]
- [x] I've **digitally signed** all my commits in this PR
- [x] I've run `make pre-commit` and confirm it didn't generate any warnings/changes
- [x] I've run `make test` and I confirm all tests passed locally
- [ ] I've added/updated any relevant documentation in the `docs/` folder 
- [x] I've linked this PR with an open issue
- [x] I've tested and verified that my code works as intended and resolves the issue as described
- [ ] I've attached screenshots demonstrating that my code works as intended (if applicable)
- [x] I've checked all other open PRs to avoid submitting duplicate work
- [x] I confirm that the code and comments in this PR are not direct unreviewed outputs of AI
- [x] I confirm that I am the Sole Responsible Author for every line of code, comment, and design decision

PR #1730, opened 2026-09-09
Fix test isolation breaking HTML report tests

### LINKED ISSUE #1721: Failed tests on FreeBSD
I'm working on adding Nettacker to the FreeBSD ports collection. Part of that was integrating the tests in tests/ into the ports framework. I ran the tests on FreeBSD 14.4-RELEASE-p9, which includes OpenSSL 3.0.20 in base.

There are some errors in the results. I've run application on the host without seeing any noticable errors, including OpenSSL tests.

[test_results.txt](https://github.com/user-attachments/files/31853970/test_results.txt)

SAI-HARISH2007:
Fixes #1721

`nettacker/lib/html_log/log_data.py` read five report templates at module import. That made import order significant. Four tests in `tests/unit/api/test_core.py` (lines 50, 57, 65, 72) assign `Config.path.web_static_dir` directly and never restore it, so a later import of `log_data` resolved against the leaked `/safe/dir` value and raised `FileNotFoundError`.

CI does not see this because the default `addopts` run the suite under xdist with `--dist loadscope`, which puts the api and database modules in separate worker processes. The leaked global never crosses between them. The failure only appears in a single-process run, which is how @einsibjarni hit it while packaging Nettacker for FreeBSD ports.

## Changes

`nettacker/lib/html_log/log_data.py` loads templates on first access through a module-level `__getattr__` and caches them, instead of reading them at import. `__all__` lists the five template names so wildcard imports still work, and the names are declared as annotations so static analysis can resolve them without binding them. Attribute access and `mock.patch` are unchanged, so `tests/unit/database/test_db.py:1081` still works.

`tests/unit/api/test_core.py` gains one autouse fixture that saves and restores `web_static_dir`. One fixture covers all four assignment sites.

`tests/unit/lib/html_log/test_log_data.py` is new. It checks that importing the module touches no files, that templates load from the current static directory, and that an unknown attribute raises `AttributeError`. Its fixture restores both the `sys.modules` entry and the parent package attribute, since `importlib` rebinds both.

## Verification

Python 3.12.13, `pytest tests/unit -o addopts=`

On master, before this branch:

    FAILED tests/unit/core/lib/test_socket.py::TestSocketMethod::test_create_tcp_socket
    FAILED tests/unit/core/lib/test_ssl.py::TestSslMethod::test_create_tcp_socket
    FAILED tests/unit/database/test_db.py::TestDatabase::test_logs_to_report_html_sqlite
    FAILED tests/unit/database/test_db.py::TestDatabase::test_logs_to_report_html_sqlalchemy
    4 failed, 338 passed, 6 skipped, 2 xfailed

On this branch, rebased onto current master:

    FAILED tests/unit/core/lib/test_socket.py::TestSocketMethod::test_create_tcp_socket
    FAILED tests/unit/core/lib/test_ssl.py::TestSslMethod::test_create_tcp_socket
    2 failed, 357 passed, 6 skipped, 2 xfailed

The passed totals differ because the rebase picked up #1707, which added tests.

Running with the default addopts gives the same 2 failed, 357 passed, so the result no longer depends on how the suite is invoked.

The two remaining failures are the `ssl.wrap_socket` tests, which #1676 covers.

`ruff check` and `ruff format --diff` are clean on the changed files. There is a pre-existing B017 in `tests/unit/api/test_core.py:98` that I left alone since it is unrelated to this fix.

This re-enables two of the three tests @einsibjarni disabled in the FreeBSD port. `test_create_tcp_socket` still depends on #1676.

### DIFF
--- a/nettacker/lib/html_log/log_data.py
+++ b/nettacker/lib/html_log/log_data.py
@@ -1,7 +1,30 @@
 from nettacker.config import Config
 
-css_1 = open(Config.path.web_static_dir / "report/html_table.css").read()
-json_parse_js = open(Config.path.web_static_dir / "report/json_parse.js").read()
-table_end = open(Config.path.web_static_dir / "report/table_end.html").read()
-table_items = open(Config.path.web_static_dir / "report/table_items.html").read()
-table_title = open(Config.path.web_static_dir / "report/table_title.html").read()
+_TEMPLATE_FILES = {
+    "css_1": "report/html_table.css",
+    "json_parse_js": "report/json_parse.js",
+    "table_end": "report/table_end.html",
+    "table_items": "report/table_items.html",
+    "table_title": "report/table_title.html",
+}
+
+
+def __getattr__(name):
+    """Read report templates on first access instead of at import time.
+
+    Reading at import made import order significant: any code that changed
+    Config.path.web_static_dir before this module was first imported caused
+    the import itself to fail.
+    """
+    try:
+        relative_path = _TEMPLATE_FILES[name]
+    except KeyError:
+        raise AttributeError(f"module {__name__!r} has no attribute {name!r}") from None
+
+    value = (Config.path.web_static_dir / relative_path).read_text()
+    globals()[name] = value
+    return value
+
+
+def __dir__():
+    return sorted([*globals(), *_TEMPLATE_FILES])
--- a/nettacker/lib/html_log/log_data.py
+++ b/nettacker/lib/html_log/log_data.py
@@ -8,6 +8,22 @@ _TEMPLATE_FILES = {
     "table_title": "report/table_title.html",
 }
 
+__all__ = [
+    "css_1",
+    "json_parse_js",
+    "table_end",
+    "table_items",
+    "table_title",
+]
+
+# Declared for static analysis and wildcard imports. These are annotations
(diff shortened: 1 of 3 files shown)


securestep9:

SAI-HARISH2007:
Thanks for the review. Addressed the three code findings in 0a02674.

`__all__` now lists the five template names, so `from log_data import *` exports them again. Module `__getattr__` alone does not, as the review noted. The names are also declared as annotations, which lets ruff resolve them without binding them and keeps the lazy loading intact.

`__dir__` returns a set union now, so a template is not listed twice once it has been accessed and cached.

The log_data tests restore the parent package attribute as well as the `sys.modules` entry. Good catch. `importlib` rebinds `nettacker.lib.html_log.log_data`, so restoring only `sys.modules` left the package pointing at a module holding templates cached from a temporary directory.

On the signing finding: the sha in that comment, 04341a95c21dde257e26fd2a14546590d26c45dd, does not exist in this repository, and the GitHub API reports `verified: true` for the branch commits. I think it inspected the ephemeral merge commit rather than the branch itself. Both commits are signed with an SSH key registered on my account.

Also rebased onto master, since the branch was two commits behind. Suite is 2 failed, 357 passed with and without xdist, and the two failures are the `ssl.wrap_socket` tests covered by #1676.

